PR #1, opened 2020-04-10
feat: first version of community health files

derberg:
With this new repo and files introduced with this PR we get global community health files, like code of conduct or issue templates, so we do not have to maintain them in each repo separately. What does it mean?
- That sponsor button will now be visible on all repos, without explicit configuration
    Screenshot 2020-04-10 at 15 46 35
- When you will create issue in a repository that has no specific issues templates, you will see this view. As you can see I suggest we allow using blank issue if users want as they might have a case with us not covered with templates. We should block this option only if we see users abuse it.
    Screenshot 2020-04-10 at 16 50 30
After merging this PR it will not yet work because of https://github.community/t5/How-to-use-Git-and-GitHub/Default-community-files-ignores-issue-templates/m-p/42584/highlight/false#M9957 but we can merge and wait for fix
- When creating issue, user will see links to all the most important community files, stored in one location, again, no need to maintain those files in all the repos
    Screenshot 2020-04-10 at 15 50 52
- Under security tab users will see info how to report security issue if they do not want to report issues publicly
    Screenshot 2020-04-10 at 15 51 41
- All repos community profiles will look like this, perfect
    Screenshot 2020-04-10 at 17 09 08

----
where some content comes from:
- code of conduct 1:1 from https://github.com/asyncapi/asyncapi/blob/master/CODE_OF_CONDUCT.md
- contribution file from https://github.com/asyncapi/asyncapi/blob/master/CONTRIBUTING.md with some improvements + info about conventional commits
- git workflow 1:1 from https://github.com/asyncapi/asyncapi/blob/master/documents/git-workflow.md
- PR and issue templates are pretty new
- security and support are new

This PR fixes https://github.com/asyncapi/asyncapi/issues/355


jonaslagoni:
Great 👍 What about adding some of the github actions is that possible 🤔 ?

derberg:
@jonaslagoni that would be awesome but unfortunately that is not supported yet. I asked them explicitly https://github.community/t5/GitHub-Actions/Plans-to-support-global-workflows-in-github-repository/m-p/50718#M7939 and then went through the suggested process to provide feedback to GitHub, I don't think though that it will be done quick

Issue #13, opened 2020-11-23
Workflow does not need to run when PR is a draft 


jonaslagoni:
#### Reason/Context
Same reasons as in https://github.com/asyncapi/parser-js/issues/200 we should disable `sentiment-analysis.yml` and `automerge.yml`. It does not make sense to run it whena PR is in draft mode.

#### Description
Add `if: github.event.pull_request.draft == false` check and ensure `ready_for_review` **on** type are added otherwise when a draft becomes ready to review it wont trigger the job.

derberg:
I think for sentiment analysis we should monitor from start, as text can be negative before it is submitted for review

PR #14, opened 2020-11-25
feat: add workflow to replicate workflows from this repo to all others

### LINKED ISSUE #12: Create custom github action that picks the change in workflow files and populates the change to all other repositories
#### Reason/Context
- GitHub Actions do not support "global" GitHub Actions (yeap, true). 
- We have many workflows that are exactly the same across all repositories. If there is a need to change something in one, we need to go to these 30-something repos and make a change, manually.....omg

#### Description

Create an action that populates changes automatically for us:
- `.github` repo is a global repo for us, with global workflows that need to be populated to all other repositories
- we change workflow, like welcome workflow, in `.github` repo and review it here
- we merge the change and action comes into play
- action picks up that change in workflow was done
  - it clones all repos
  - updates the same workflow file in every repo
  - creates new branch with change and creates pull request
  - automerge workflow in every repo automatically approves and merges PR as it comes from asyncapi-bot
- puppies get crazy of happiness 
Please try answering few of those questions
- What changes have to be introduced?
- Will this be a breaking change?
- How could it be implemented/designed?

derberg:
**Description**

- add workflow to replicate workflows from this repo to all others
- with this action, doing https://github.com/asyncapi/.github/issues/13 will be as easy as just editing only https://github.com/asyncapi/.github/blob/master/.github/workflows/automerge.yml and this file will be replicated in all the other repositories
- action source code https://github.com/marketplace/actions/global-workflows-support

**Related issue(s)**
Resolves #12

### DIFF
--- a/.github/workflows/global-workflows-support.yml
+++ b/.github/workflows/global-workflows-support.yml
@@ -1,0 +1,27 @@
+name: Global workflow to rule them all
+
+on:
+  push:
+      branches: [ master ]
+
+jobs:
+
+  replicate_changes:
+
+      runs-on: ubuntu-latest
+
+      steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Replicating global workflows
+        uses: derberg/global-workflows-support@v0.0.2
+        with:
+          github_token: ${{ secrets.GH_TOKEN }}
+          # this action will not replicate to other repos workflows listed here
+          files_to_ignore: global-workflows-support.yml
+          # workflows will not be replicated to repos listed here
+          repos_to_ignore: shape-up-process
+          committer_username: asyncapi-bot
+          committer_email: info@asyncapi.io
+          # it is both, commit message and PR title
+          commit_message: "ci: update global workflows"


.github/workflows/global-workflows-support.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Replicating global workflows
+        uses: derberg/global-workflows-support@v0.0.2
+        with:
+          github_token: ${{ secrets.GH_TOKEN }}
+          # this action will not replicate to other repos workflows listed here
+          files_to_ignore: global-workflows-support.yml
+          # workflows will not be replicated to repos listed here
+          repos_to_ignore: shape-up-process
```
fmvilas, line 23: Shouldn't this be an array?

derberg: should, but GH Actions do not support input other than string yet 😞  this is why comma separated list here

fmvilas (approved):

PR #18, opened 2021-01-11
ci: add new global workflow for slack and twitter notifications

derberg:
**Description**

- New workflow that will be added to all repositories for sending notifications to slack on new PRs and Issues
  - separate channel #github-new-issues-prs
  - dependabot and asyncapi bot filtered out
  - example messages you can already see in #github-new-issues-prs channel
- New workflow responsible for releases announcements
 - all releases from all repos will go to slack to #github-releases (example already there)
 - Twitter (as agreed, only for minor and major) will be integrated into all repos. How we check if there is a new release, and what type of release is it was improved as package.json or npm scripts are no longer needed. This solution works now with all types of repos that do GitHub Releases

as a result of this PR:
- we should remove github-activity channel and app integration  
- before merging I need to adjust release workflow for HTML template and parser as we already have Twitter there

a proof that it works is that they were tested in asyncapi-react component. Results:
- https://asyncapi.slack.com/archives/C01JF00UGKC/p1610967420000100
- https://asyncapi.slack.com/archives/C01J06RL10X/p1610733591001100
- https://twitter.com/AsyncAPISpec/status/1350030328623214594

**Related issue(s)**


derberg:
there is one job failing as there was a webhook missing in secrets. Now it is there and tbh I tried to restart this job, even by creating new prs but it doesn't seem to pick up changes from secrets that I did after this job failed

derberg:
this is what happends to release workflow after removing twitter from it -> https://github.com/asyncapi/html-template/pull/125

fmvilas (approved):

PR #21, opened 2021-01-12
ci: add generic version bump and pr testing and relese workflows for nodejs

derberg:
**Description**

- these are workflows that will be populated to all repos, even non-nodejs, but will only work if `package.json` is in the root

**version bump**
- the biggest advantage of this one is that PR creation action that we depend on will be updated globally only in this workflow (easier to fix issues like https://github.com/asyncapi/markdown-template/issues/55), and will not be an integral part of individual release workflows
- we will have to maintain 2 scripts in every nodejs project:
  - `generate:assets` that will run scripts like the ones about generating docs. toc in readme or types
  - `bump:version` that will run scripts to bump version of the package in package.json

**pr testing**
- changing what nodejs version we are testing in all repos will be as simple as updating just this one single workflow
- we introduce matrix testing on not only linux but also macos and windows

**release workflow**
This is what I'm happy with the most, that was not my intention at the beginning but after testing it with React component that had the most complicated release workflow I'm now sure that we can have a simple generic release workflow that can be easily extended per repo with the approach we did in react https://github.com/asyncapi/asyncapi-react/blob/master/.github/workflows/release-wc-and-playground.yml

**Related issue(s)**

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Check if Node.js project and has package.json
         id: packagejson
-        run: test -e ./package.json && "::set-output name=exists::true" || "::set-output name=exists::false"
+        run: test -e ./package.json && echo ::set-output name=exists::true || echo ::set-output name=exists::false
       - if: steps.packagejson.outputs.exists == 'true'
         name: Assets generation
         run: npm run generate:assets
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -20,14 +20,13 @@ jobs:
         run: test -e ./package.json && "::set-output name=exists::true" || "::set-output name=exists::false"
       - if: steps.packagejson.outputs.exists == 'true'
         name: Assets generation
-        run: npm generate:assets
+        run: npm run generate:assets
       - if: steps.packagejson.outputs.exists == 'true'  
         name: Bump version in package.json
         # There is no need to substract "v" from the tag as version script handles it
         # We make sure here no tags are added by default as they are already added by release workflow
         # --allow-same-version is set in case someone forgot and updated package.json manually and we want to avoide this action to fail and raise confusion
-        # run: npm --no-git-tag-version --allow-same-version version ${{github.event.release.tag_name}}
-        run: VERSION=${{github.event.release.tag_name}} npm bump:version
+        run: VERSION=${{github.event.release.tag_name}} npm run bump:version
       - if: steps.packagejson.outputs.exists == 'true'
         name: Create Pull Request with updated asset files including package.json
         uses: peter-evans/create-pull-request@v3
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -1,0 +1,41 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+#It does magic only if there is package.json file in the root of the project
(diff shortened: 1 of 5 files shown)


.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,41 @@
...
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Assets generation
+        run: npm generate:assets
+      - if: steps.packagejson.outputs.exists == 'true'  
+        name: Bump version in package.json
+        # There is no need to substract "v" from the tag as version script handles it
+        # We make sure here no tags are added by default as they are already added by release workflow
+        # --allow-same-version is set in case someone forgot and updated package.json manually and we want to avoide this action to fail and raise confusion
+        # run: npm --no-git-tag-version --allow-same-version version ${{github.event.release.tag_name}}
+        run: VERSION=${{github.event.release.tag_name}} npm bump:version
```
derberg, line 30: it is done this way, with exporting version as environment variable as it is easier this way to later implement `npm bump:version` script in monorepos

derberg:
this is what happens to release workflow after removing bump and or creation from it -> https://github.com/asyncapi/html-template/pull/125

jonaslagoni (commented):
When this is approved, how will it handle templates which does not contain these new scripts?

derberg:
@jonaslagoni it won't handle them gracefully. The job will fail. But it is not a problem as it won't happen 😄 
Once this is approved, I know it is a direction we want to take and I have to first go to all repos with the current release pipeline and adjust it:

- add/adjust scripts
- remove bump and or creation from releases
- etc

So this one is not merged until PRs from all the other repos are merged.
I also want to add all those new "global" workflows manually to react component, and make adjustments there and run real release there to make sure all those ideas work as expected

jonaslagoni (approved):
LGTM 👍

derberg:
@jonaslagoni awesome. What do you think about taking the same approach (generic nodejs workflow with first step that checks if package.json is there) for PR testing and release workflow in general -> https://github.com/asyncapi/html-template/pull/125/files#r555705827

jonaslagoni:
> @jonaslagoni awesome. What do you think about taking the same approach (generic nodejs workflow with first step that checks if package.json is there) for PR testing and release workflow in general -> https://github.com/asyncapi/html-template/pull/125/files#r555705827

@derberg I dont have any objections in standardize the release process for npm repositories 👍

derberg:
@jonaslagoni I'll update this PR then, let's see how it will look like

derberg:
ok, now this PR has 3 new global workflows that will run on nodejs projects only.

I don't like this:
```
        name: Setup Node.js
        uses: actions/setup-node@v1
        with:
          node-version: 14
```

but there is no alternative now, spent some more hours on investigation. Light in the tunnel for the future is that we could just  do the following condition on a job level `if: contains(github.event.repository.topics, 'nodejs')` and decide that only repo with such topic will apply such a workflow (which is handy) but this is not going to happen soon as `topics` (even though old feature) are not yet part of `repository` event  😱  -> https://github.community/t/why-topics-are-not-part-of-repository/155130

derberg:
@jonaslagoni I updated this PR with generic workflows for release and pr testing. As a result, this is how it will look like in other projects once this PR is merged -> https://github.com/asyncapi/html-template/pull/125/files

Looks promising but as mentioned before now I will do the cleanup in a most complex project, react component

derberg:
@jonaslagoni @magicmatatjahu I updated this PR, including the description. I think it is awesome that we can have generic release workflow for all nodejs projects. All workflows tested against asyncapi-react with success

Of course, this is not going to be merged once I cleanup release workflows across all repos

fmvilas (approved):

PR #22, opened 2021-01-19
ci: add monitoring if there are not merged PRs that should be automerged

derberg:
**Description**

- As a result of this PR, all repos will get a workflow that runs once a day at midnight and reports if there are any PRs from asyncapi-bot or dependabot that are not merged, while they should

This is what message will show up in #github-new-issues-prs channel
Screenshot 2021-01-19 at 19 02 22

Why we need this? @jonaslagoni found this https://github.com/asyncapi/raml-dt-schema-parser/pull/25 and we agreed that next time we should definitely learn about those cases in an automated way. This PR is a proposal from my side on how we can do this

### DIFF
--- a/.github/workflows/automerge-orphans.yml
+++ b/.github/workflows/automerge-orphans.yml
@@ -1,0 +1,63 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+name: 'Notify on failing automerge'
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  identify-orphans:
+    name: Find orphans and notify
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get list of orphans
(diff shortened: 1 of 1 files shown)


fmvilas (approved):

jonaslagoni:
Just a quick question, why not tag/assign one of us in the PR thats failing so that out toast messenger app in slack will take over?

derberg:
@jonaslagoni but who? now notification would go to a channel that all watch

jonaslagoni:
You could just tag all of us 😄 I would hate to have more channels, but thats just my opinion 😄

jonaslagoni:
or use @ asyncapi/maintainers

derberg:
@jonaslagoni just for the record, it is not a new channel, that would be #github-new-issues-prs channel that we anyway agreed on having, next to #github-releases as a replacement for #github-activity

derberg:
at the end, you should not be bothered during cycle, the current maintainer should have a look, and this person already should watch  #github-new-issues-prs channel

jonaslagoni:
👍

PR #23, opened 2021-01-19
ci: add pull request title linter workflow

derberg:
**Description**

- We discussed we need PR title linter before we can think of automating PR merges through `/merge` comment, instead of asking people to click merge button
- My proposal is to use `amannn/action-semantic-pull-request` as it is the only action that also gives an option to validate the uppercase/lowercase of the subject. I contributed there last week a way to provide a custom error message that explains what went wrong with validation in a more human-friendly manner. 
- yes, we never talked about `should it be uppercase` or `should it be lowercase`. I just noticed I do lowercase always but the majority do uppercase. So let us just stick to one

**Related issue(s)**

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -1,0 +1,21 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+name: Lint PR title
+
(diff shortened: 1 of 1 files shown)


fmvilas (approved):

derberg:
sorry, change of plans, we need to actually lowercase to be consistent as this is what allcontributorsbot and dependabot uses 🤷🏼‍♂️

fmvilas (approved):

PR #42, opened 2021-03-24
ci: add go-pr-testing workflow

smoya:
Part of https://github.com/asyncapi/shape-up-process/issues/94
This PR adds a new workflow for **Go** projects. It runs two steps:

- **lint**
  - It runs [golangci-lint](https://golangci-lint.run/) linter, which it is a tool that runs the most common linters for go.
  - Note: The configuration for the linters are up to each repository (via `.golangci.yml` file), however I think it makes sense if we create a repository template for Go projects that contain, among others, that file. I created a [gist](https://gist.github.com/smoya/35720dd325025dc5665a4f86d1849aae) containing all the config we might want to set.
- **test**
  - It ensures the app can be built.
  - It runs tests.

Note: The golangci-lint gh action not only fails the build in case of lint issues but it shows those issues in the PR review as a check runs. 
Example:
image

### DIFF
--- a/.github/workflows/if-go-pr-testing.yml
+++ b/.github/workflows/if-go-pr-testing.yml
@@ -1,6 +1,6 @@
 #This action is centrally managed in https://github.com/asyncapi/.github/
 #Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
-#It does magic only if there is package.json file in the root of the project
+#It does magic only if there is go.mod file in the root of the project
 name: PR testing - if Go project
 
 on:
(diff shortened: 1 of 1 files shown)


jonaslagoni (dismissed):
Looks fined to me, but we might need to start consider another approach in regards to sharing global workflow for multiple languages 🤔 In the end it could be a lot of checks 😅

.github/workflows/if-go-pr-testing.yml
```diff
@@ -0,0 +1,58 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+#It does magic only if there is package.json file in the root of the project
```
derberg, line 3: go.mod please and not package.json

derberg (changes requested):
Once you make a small fix I approve as all is good.

.github/workflows/if-go-pr-testing.yml
smoya: Fixed!

derberg (approved):

PR #43, opened 2021-04-01
ci: update autoapprove workflow to v2 and pr target event

derberg:
**Description**

- recently autoapprove workflows were failing and this is the issue https://github.com/hmarr/auto-approve-action/issues/183 that explains what has changed to default GITHUB token

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -4,7 +4,7 @@
 name: Automerge release bump PR
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - labeled
       - unlabeled
@@ -17,10 +17,6 @@ on:
   pull_request_review:
     types:
       - submitted
-  check_suite: 
-    types:
-      - completed
-  status: {}
   
 jobs:
 
@@ -29,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Autoapproving
-        uses: hmarr/auto-approve-action@v2.0.0
+        uses: hmarr/auto-approve-action@v2
         if: github.actor == 'asyncapi-bot' || github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"


.github/workflows/automerge.yml
```diff
@@ -17,10 +17,6 @@ on:
   pull_request_review:
     types:
       - submitted
-  check_suite:
```
derberg, line 20: I don't think we even needed to react to those events in our workflows

.github/workflows/automerge.yml
```diff
@@ -29,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Autoapproving
-        uses: hmarr/auto-approve-action@v2.0.0
+        uses: hmarr/auto-approve-action@v2
```
derberg, line 28: no we will not have to bump to every new 2.x version but always get latest improvements, line now we will also use 2.1 that improves logging of errors

magicmatatjahu: We don't need to tell like in npm `^v2` or something like that?

.github/workflows/automerge.yml
```diff
@@ -4,7 +4,7 @@
 name: Automerge release bump PR
 
 on:
-  pull_request:
+  pull_request_target:
```
magicmatatjahu, line 7: Why do you wanna change it? I only ask 😄

.github/workflows/automerge.yml
derberg: nope, it is all fetched from GitHub at the end, super manual process for action owner. You release for example v2.1.0 and then have to update release v2 again to match v2.1.0 🤷🏼‍♂️

.github/workflows/automerge.yml
derberg: explanation is in linked issue. The problem is that in march github changed the scope of default GITHUB_TOKEN and it has read only rights on `pull_request` event, so the action fails as it has no rights to updat the status of the check through the API 😞 `pull_request_target` is relatively new event created mainly for GH Actions, so you basically mark that actions runs in scope of base not fork.

I'm a proud fighter for `pull_request_target` https://dev.to/derberg/github-actions-when-fascination-turns-into-disappointment-4d75 😄

magicmatatjahu: Thanks for explanation!

magicmatatjahu (approved):

PR #44, opened 2021-04-06
ci: bump automerge action to version supporting pull_request_target

derberg:
**Description**

- This change is needed as my last change made the workflow react on `pull_request_target` to fix `autoapprove` step, but it broke `automerge` that in past versions did not support new event 😞


derberg:
I removed roadmap repo so no need to ignore it anymore, we didn't need this repo as roadmap is part of shapeup repo

magicmatatjahu (approved):

PR #45, opened 2021-04-15
ci: allow running global-workflows-support workflow manually

smoya:
This PR change allows triggering `global-workflows-support` workflow manually from Github UI. 
More info on https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/global-workflows-support.yml
+++ b/.github/workflows/global-workflows-support.yml
@@ -3,6 +3,7 @@ name: Global workflow to rule them all
 on:
   push:
       branches: [ master ]
+  workflow_dispatch:
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/global-workflows-support.yml
```diff
@@ -3,6 +3,7 @@ name: Global workflow to rule them all
 on:
   push:
       branches: [ master ]
+  workflow_dispatch:
```
derberg, line 6: are you sure it is correct? for sure this is not how you do empty object in yaml?

shouldn't it me just `workflow_dispatch` or `workflow_dispatch: {}`

smoya: Yeah, that will mean `null`. I'm gonna move it to `workflow_dispatch: {}`

derberg (approved):

PR #46, opened 2021-04-19
ci: bump global workflow to support workflow_dispatch event

derberg:


magicmatatjahu (approved):

derberg:
@smoya I had to do some more coding to enable that 😄 
https://github.com/derberg/global-workflows-support/pull/15/files

derberg:
@smoya it works like a charm 🍻 https://github.com/asyncapi/template-for-go-projects/pull/2/files

it went through all repos but only 4 PR were created https://github.com/asyncapi/.github/runs/2378677229?check_suite_focus=true which is awesome.

can you please update https://github.com/asyncapi/.github/blob/master/repository-settings.md with everything that you had to do with new repo but was not described? + write that instead of replicating workflow files manually, you can not go and run replication manually with one click 🎆

PR #48, opened 2021-04-19
docs: update how to install global workflows

smoya:
This PR updates the instructions for installing global workflows, which can now be installed through GH actions.

### DIFF
--- a/repository-settings.md
+++ b/repository-settings.md
@@ -34,5 +34,5 @@ AsyncAPI Github organization [global workflows](https://github.com/asyncapi/.git
 * Release workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.ymll)
 
 ### Installation
-Run **manually** the [Global workflow to rule them all](https://github.com/asyncapi/.github/actions/workflows/global-workflows-support.yml) workflow as shown in the following screenshot:
+Add basic workflow by **manually** triggering the [Global workflow to rule them all](https://github.com/asyncapi/.github/actions/workflows/global-workflows-support.yml) workflow as shown in the following screenshot:
 ![image](https://user-images.githubusercontent.com/1083296/115218253-a5e42480-a106-11eb-9723-165b9ba90e93.png)
--- a/repository-settings.md
+++ b/repository-settings.md
@@ -22,6 +22,7 @@ Leave only the **Allow squash merging** option selected. This option combines al
 ## SonarCloud scans
 
 Each repository must be integrated with https://sonarcloud.io/organizations/asyncapi/projects for automated quality and security scans.
+If your project is not in the list, please add it via https://sonarcloud.io/projects/create.
 
 ## Add basic GitHub Actions configurations
 
--- a/repository-settings.md
+++ b/repository-settings.md
@@ -33,6 +33,5 @@ AsyncAPI Github organization [global workflows](https://github.com/asyncapi/.git
 * Automated bump of the npm package in its dependent project. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-version-bump.yml)
 * Release workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.ymll)
 
-### Installation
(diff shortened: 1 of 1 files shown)


repository-settings.md
```diff
@@ -25,11 +25,14 @@ Each repository must be integrated with https://sonarcloud.io/organizations/asyn
...
-* Release workflow that should be inspired by [this](https://github.com/asyncapi/parser-js/blob/master/.github/workflows/release.yml)
\ No newline at end of file
+AsyncAPI Github organization [global workflows](https://github.com/asyncapi/.github/tree/master/.github/workflows) cover mostly all the needs. For example:
+* Handling of stale issues and PRs. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/stale-issues-prs.yml)
+* Welcome message for first-time contributors. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/welcome-first-time-contrib.yml)
+* Automerge workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/automerge.yml)
+* Sentiment analysis workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/sentiment-analysis.yml)
+* Automated bump of the npm package in its dependent project. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-version-bump.yml)
+* Release workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.ymll)
```
jonaslagoni, line 34: What if we split the description of the workflow up into something like:
```suggestion
For Node.js projects we have:
* Automated bump of the npm package in its dependent project. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-version-bump.yml)
* Release workflow. [file](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.yml)
```

Since we have different workflows dependent on the language, this seems fitting 🤔

jonaslagoni (commented):
Just a small suggestion 👍

repository-settings.md
```diff
@@ -25,11 +25,14 @@ Each repository must be integrated with https://sonarcloud.io/organizations/asyn
...
+
+### Installation
```
derberg, line 36: I don't think we need a section called installation if the previous already has `Add` word. Just keep below text as part of `Add basic GitHub Actions configurations`

repository-settings.md
```diff
@@ -25,11 +25,14 @@ Each repository must be integrated with https://sonarcloud.io/organizations/asyn
...
+Run **manually** the [Global workflow to rule them all](https://github.com/asyncapi/.github/actions/workflows/global-workflows-support.yml) workflow as shown in the following screenshot:
```
derberg, line 37:
```suggestion
Add basic workflow by **manually** triggering the [Global workflow to rule them all](https://github.com/asyncapi/.github/actions/workflows/global-workflows-support.yml) workflow as shown in the following screenshot:
```

derberg (changes requested):
@jonaslagoni you also added a new repo recently, you think anything more needs to be updated in this instruction, anything missing?

repository-settings.md
smoya: I vote for removing that paragraph since it's adding debt since the first moment is just talking about a possible outdated version of this: https://github.com/asyncapi/.github/tree/master/.github/workflows. I think the name on the workflows makes it pretty clear.

repository-settings.md
smoya: done

repository-settings.md
derberg: tbh, I think we can remove this first part of the section entirely and just replace with simple sentence that there are default workflows for all repositories and link to workflows dir 😄 this list can easily get outdated, links broken, etc

smoya: I think we mean the same here. I removed it.

derberg (approved):
@smoya side comment from my side - I take back my bashing of `workflow_dispatch` solution. Now I have to say I actually love it because it allows custom input. This means we can have in figure some options in global workflow that when you manually trigger it, you specify the name of the repo that workflow should be triggered against. Cool stuff

PR #50, opened 2021-04-22
ci: exclude parser-api from global workflows

smoya:
This PR adds https://github.com/asyncapi/parser-api repository to the ignored repositories list so global workflows are not installed.
The reason is that https://github.com/asyncapi/parser-api repository will only contain documentation (at least by now).


derberg:
global repos are not only about pipelines, they also have workflows for stale issues, for slack alerts when new issue or PR is created, etc

smoya:
Oh yeah, good point! Closing🙌

PR #52, opened 2021-05-19
ci: pin to nodejs workflows next and releases branches

magicmatatjahu:
**Description**

- Pin `next` and `**-release` branches to `if-nodejs-release` workflows.
- Update `if-nodejs-version-bump` workflow including `next` branch.

`next` branch at the moment is only used in `react-component` repo.

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -7,6 +7,8 @@ on:
   push:
     branches:
       - master
+      - next
+      - '**-release'
 
 jobs:
 
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -7,6 +7,8 @@ on:
   push:
     branches:
       - master
+      # below lines are not enough to have release supported for these branches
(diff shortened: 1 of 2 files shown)


derberg (commented):
Just revert changes you have done to ` .github/workflows/if-nodejs-version-bump.yml` and only modify line `ref: master` to below

```
          # target branch of release. More info https://docs.github.com/en/rest/reference/repos#releases
          ref: ${{ github.event.release.target_commitish }} #in case release is created from release branch then we need to checkout from given branch
```

.github/workflows/if-nodejs-release.yml
```diff
@@ -7,6 +7,8 @@ on:
   push:
     branches:
       - master
+      - next
```
derberg, line 10:
```suggestion
      #below lines are not enough to have release supported for these branches
      #make sure configuration of `semantic-release` package mentiones these branches
      - next
```

derberg (approved):
lgtm 🚀

magicmatatjahu:
@derberg Thanks! 🚀

Issue #34, opened 2021-02-05
Bump workflow needs to be global too, even if used in few repos only


derberg:
#### Reason/Context
The action that is used there is not yet very well battle-tested and requires updates.
For example, now it doesn't work on our org if `ts-nats-template` is processed.
We should just bump https://github.com/derberg/org-projects-dependency-manager in only one global workflow and not in several workflows in different repos

#### Description
- create here `bump.yml` file (later it will be updated in other repos where it already resides)

PR #67, opened 2021-08-11
feat: introduce bump workflow as global

derberg:
**Description**

- merge of this PR cause distribution of this workflow in all repositories and overwrite of the following files in the following repos:
   - https://github.com/asyncapi/generator/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/generator-hooks/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/generator-react-sdk/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/modelina/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/raml-dt-schema-parser/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/openapi-schema-parser/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/parser-js/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/generator-filters/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/spec-json-schemas/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/converter-js/blob/master/.github/workflows/bump.yml
   - https://github.com/asyncapi/avro-schema-parser/blob/master/.github/workflows/bump.yml
   - react-component has bump.yaml and not bump.yml so it would not be overwritten + it is not on master branch. @magicmatatjahu can you handle rename in your branch when you will make your next or?
- thanks to making bump global it will be distributed not only to HTML and markdown templates that do not have it (thus no automated bumps in playground) but also all new repos, like diff or optimizer that in future will be used in CLI and Studio
- using v2 of dependency manager so default commit messages are not part of the action thus we do not need to add custom scripts in our projects to get package name and version - simplified workflow description

do you think it is required to go to all repos mentioned above and remove these scripts:
```
    "get:version": "echo $npm_package_version",
    "get:name": "echo $npm_package_name",
```
I kinda do not feel like this is super important, they are not needed but also not harmful, but I'm opened for arguments to go and remove them in all the repositories

**Related issue(s)**
Resolves https://github.com/asyncapi/.github/issues/34

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -24,8 +24,9 @@ jobs:
         run: test -e ./package.json && echo "::set-output name=exists::true" || echo "::set-output name=exists::false"
       - if: steps.packagejson.outputs.exists == 'true' && startsWith(github.event.commits[0].message, 'chore(release):')
         name: Bumping latest version of this package in other repositories
-        uses: derberg/org-projects-dependency-manager@v2
+        uses: derberg/custom-dependabot-for-your-github-org@v2
         with:
           github_token: ${{ secrets.GH_TOKEN }}
           committer_username: asyncapi-bot
           committer_email: info@asyncapi.io
+          repos_to_ignore: html-template #this is temporary until react component releases 1.0, then it can be removed
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -1,0 +1,31 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+
+#Purpose of this action is to update npm package in libraries that use it. It is like dependabot for asyncapi npm modules only. 
+#It runs in a repo after merge of release commit and searches for other packages that use released package. Every found package gets updated with lates version
+
+name: Bump package version in dependent repos - if Node project
+
+on:
+  #It cannot run on release event as when release is created then version is not yet bumped in package.json
+  #This means we cannot extract easily latest version and have a risk that package is not yet on npm
+  push:
+    branches:
+      - master
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+      - name: Check if Node.js project and has package.json
+        id: packagejson
+        run: test -e ./package.json && echo "::set-output name=exists::true" || echo "::set-output name=exists::false"
+      - if: steps.packagejson.outputs.exists == 'true' && startsWith(github.event.commits[0].message, 'chore(release):')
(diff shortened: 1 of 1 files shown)


fmvilas (approved):
Super clean 👏

magicmatatjahu:
> react-component has bump.yaml and not bump.yml so it would not be overwritten + it is not on master branch. @magicmatatjahu can you handle rename in your branch when you will make your next or?

@derberg It's not a good idea to have the `bump.yaml` file in the master of `react-component` - why? Bumping version in the master will propagate the changes to the html-template and then we will have the old version in the template... for that we have the `bump` only in the `next` branch. Or maybe I'm wrong in this case and it won't be a problem? 🤔 

> I kinda do not feel like this is super important, they are not needed but also not harmful, but I'm opened for arguments to go and remove them in all the repositories

We can remove them as followups, each person which is "responsible" for given repos :)

derberg:
> Bumping version in the master will propagate the changes to the html-template and then we will have the old version in the template

@magicmatatjahu yes, this is how it works, as in `html-template` we do not do pre-releases with next react component.
Maybe it is time to release 1.0 of react component 😆 

🤔 
the only option I see is set HTML template as ignored, but then we need to know that any updates from asyncapi packages will not get automatically into the HTML template

derberg:
@magicmatatjahu have a look at my change. Of course your bump in next branch will still work and update html-template automatically

magicmatatjahu:
> the only option I see is set HTML template as ignored, but then we need to know that any updates from asyncapi packages will not get automatically into the HTML template

Maybe we should ignore the `react-component` and update bump manually in the `next` branch until the `1.0.0` version? It's better option because HTML template also uses the parser under the hood.

derberg:
> Maybe we should ignore the react-component and update bump manually in the next branch until the 1.0.0 version? It's better option because HTML template also uses the parser under the hood.

it is not possible to ignore global workflows update for a repo only for single workflow, you can ignore for all updates. I think it would be worse to exclude react component from workflows update rather than exclude html-template from dependencies updates, given that parser is not actively developed

magicmatatjahu:
Ok then, we can ignore html-template :)

> it is not possible to ignore global workflows update for a repo only for single workflow

It sounds as  good feature 😄

magicmatatjahu (approved):

derberg:
> It sounds as good feature 😄

it does, I was thinking about it for some time already, go ahead and kick of discussion, for now I do not see any other option as providing a list of workflow files and list of repos to ignore - a string of `|` separated key (filename) and value (list of repos comma separated) split with `:`. Other options - some config file instead of such encrypted string :)

derberg:
@magicmatatjahu @fmvilas ok folks, I also cleaned up not needed scripts, wanted to have some fun with new GitHub mindblowing DOT feature 😆

Issue #59, opened 2021-06-29
Alot of umerged PRs


jonaslagoni:
#### Describe the bug
At the moment with the CI system, it can occur that we are ended up with a lot of unmerged PRs as such:
![billede](https://user-images.githubusercontent.com/13396189/123793577-27fc8080-d8e2-11eb-9121-acad85183282.png)

#### How to Reproduce
I have no idea, it happens from time to time 😄 

#### Expected behavior
I would expect the CI system to handle this semi-automatically or give us notice of unmerged PR's as it does in some cases but not all.

derberg:
@jonaslagoni for now no idea what is the reason, we need more logs. Please approve https://github.com/asyncapi/.github/pull/66

derberg:
@jonaslagoni figured out what was wrong, have a look at a pr

PR #69, opened 2021-08-12
fix: improve alerts on orphan bot prs and discussions and use latest bump

derberg:
**Description**

## Orphans

There was a problem the workflow that checks daily for unmerged PRs was not always reporting all the PRs. Some PRs were really waiting for ages to be merged.

Reason was that workflow is not paging-proof and in repos that have over 100PRs it could never work. Now GraphQL query pulls only PRs that have state `OPEN`. So now this action is future-proof and we will have to start to worry if fails only when we will have more than 100 open PRs on a repo. I think once that happens we will have many other bigger issues than this 😆 

**Related issue(s)**
Fixes https://github.com/asyncapi/.github/issues/59

## Discussion

Since discussions are out of beta, now GitHub Actions also support it and we should integrate them with our existing workflow that notifies us about every new pr and issue

## Bump workflow

Had to release another version of bump-responsible action, there was a bug fix + there was also a legal problem I had with GitHub because old name of action had "dependabot" in it 🤷🏼 , now it is called `npm-dependency-manager-for-your-github-org`/

### DIFF
--- a/.github/workflows/automerge-orphans.yml
+++ b/.github/workflows/automerge-orphans.yml
@@ -37,9 +37,7 @@ jobs:
             };
             const { repository: { pullRequests: { nodes } } } = await github.graphql(query, variables);
 
-            let orphans = nodes.filter((pr)=> {
-              pr.author.resourcePath === '/asyncapi-bot' || pr.author.resourcePath === '/apps/dependabot'
-            })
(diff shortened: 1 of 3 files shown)


jonaslagoni (approved):
👍

jonaslagoni (approved):

jonaslagoni:
> @jonaslagoni yo, one last change added to this PR. Details in description in `Bump workflow` section. Once you approve we merge

Are you sure there are no other changes? 😆

jonaslagoni (approved):

PR #75, opened 2021-09-22
chore: bring welcome workflow back but disabled

derberg:
global workflow action doesn't support yet removal of workflows, only when they are created or modified. So need to bring workflow back and disable it, only this way it will be propagated to other repos


magicmatatjahu (approved):
😅

PR #77, opened 2021-09-27
ci: increase time on stale bot reaction and use latest version

derberg:
**Description**

- increase stale action to 120days. So issue is staled after 120days and then closed after 120, so it is 240 days before issues get closed.
- switch from v1 to v4 of stale action
- new version of stale action support removal of `stale` label after adding a comment (before you had to manually remove label)
- added more detailed message to stale action comment

The reason for the change is the increasing amount of issues/prs coming from the community, but the active contributors group is not growing at the same speed. We are not able to react to every single issue quickly enough to not get spammed by stale bot.

⚡ Important to notice ⚡ 
With this change, we can "unstale" issues with a comment, no need to manually remove the label. As a result, every maintainer should always favor the process of "unstaling by commenting" rather than "unstaling by removing the `stale` label".  Why? because this forces you (maintainer) to think why a given issue got stale, and also give hints on how to push a given issue forward. It is all about transparency and sending a clear messages to the community rather than expecting people to figure out the reason on their own.


smoya (approved):
Love it! Thanks

jonaslagoni (approved):
Sounds good to me 👍

magicmatatjahu (approved):
LGTM! 🚀

fmvilas (approved):
Awesome! 🚀

PR #84, opened 2021-10-01
ci: update to latest Setup Node.js workflow to use cache

### LINKED ISSUE #80: workflows that require `npm install` are long running and sometimes blocking
We test all npm packages on macos, windows and linux. Macos is the slowest and sometimes `npm install` fails. Sometimes you don't really want to wait 15min for a PR test to complete.

Hint:

There are some solutions out there that can be plugged into the workflow to cache npm modules or just reuse some other cache.

HollowMan6:
**Description**
- https://github.com/marketplace/actions/setup-node-js-environment?version=v2.4.1#caching-packages-dependencies

**Related issue(s)**
Hope to resolve #80

Signed-off-by: Hollow Man <hollowman@hollowman.ml>

### DIFF
--- a/.github/workflows/if-nodejs-pr-testing.yml
+++ b/.github/workflows/if-nodejs-pr-testing.yml
@@ -24,7 +24,7 @@ jobs:
         shell: bash
       - if: steps.packagejson.outputs.exists == 'true'
         name: Setup Node.js
-        uses: actions/setup-node@v2.4.1
+        uses: actions/setup-node@v2
         with:
           node-version: 14
           cache: 'npm'
--- a/.github/workflows/if-nodejs-pr-testing.yml
+++ b/.github/workflows/if-nodejs-pr-testing.yml
@@ -24,9 +24,11 @@ jobs:
         shell: bash
       - if: steps.packagejson.outputs.exists == 'true'
         name: Setup Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2.4.1
         with:
           node-version: 14
+          cache: 'npm'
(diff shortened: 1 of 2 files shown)


HollowMan6:
Tested cache npm modules using the same code
https://github.com/openSUSE/Customize-IBus/blob/9260fa4cdfa1a5b8559a60e9e988283516668287/.github/workflows/npm-lint.yml#L18-L22

**Result**:
https://github.com/openSUSE/Customize-IBus/runs/3767167054?check_suite_focus=true#step:3:15

.github/workflows/if-nodejs-pr-testing.yml
```diff
@@ -24,9 +24,11 @@ jobs:
         shell: bash
       - if: steps.packagejson.outputs.exists == 'true'
         name: Setup Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2.4.1
```
derberg, line 27: please change to `actions/setup-node@v2` so we always use latest, it is recommended best practice from GitHub so we do not need to update every time some bug fix is published for v2

derberg (commented):
@HollowMan6 thanks so much for taking time to work on this!!! I made only small suggestion to fix usage on `v2` and not specific patch release

HollowMan6:
@derberg Done! Thanks for reviewing!

derberg (approved):
@HollowMan6 thanks a lot!

fyi there are more issues like this, just check out [this](https://github.com/asyncapi/.github/issues?q=is%3Aissue+is%3Aopen+label%3AHacktoberfest) list

Issue #68, opened 2021-08-12
Automerge should auto merge master


derberg:
Quite often automerge fails because like [here](https://github.com/asyncapi/generator-filters/pull/50) PR was created, but in the meantime something else was merged and therefore PR was out of sync with master - thus blocked from merging.

https://github.com/asyncapi/.github/blob/master/.github/workflows/automerge.yml workflow needs a step that updates PR to latest master. 

We need integrate something like https://github.com/marketplace/actions/auto-update

abxhr:
I would love to work on this! 🚀

derberg:
Please go ahead and lemme know if ya need help

derberg:
This one is already fixed

PR #85, opened 2021-10-01
feat: add autoupdate in workflow

abxhr:
**Description**

This PR adds an autoupdate step for pull request GitHub action, which updates the branch for the PR, if any changes are made in the `master` branch.

**Related issue(s)**

**Fixes** #68


.github/workflows/automerge.yml
```diff
@@ -17,9 +17,23 @@ on:
   pull_request_review:
     types:
       - submitted
+  push:
```
derberg, line 20: why do we need that? this workflow is for PRs only

.github/workflows/automerge.yml
```diff
@@ -17,9 +17,23 @@ on:
...
 
+  autoupdate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Autoupdating
+        uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+          PR_FILTER: "labelled"
+          PR_LABELS: "autoupdate"
```
derberg, line 34: I don't recall us having such a label, who can it help? Do you think we should run this only on the PR handled by the `asyncapi-bot`? I was thinking that maybe on all 🤔 can't find a reasons why wouldn't we 🤔

derberg (commented):
@abxhr oh man, thanks so much. you can't imagine how helpful this is for us. I left some questions to understand what you are proposing before we merge

also, was thinking that we need to set `MERGE_CONFLICT_ACTION` to `ignore` because conflicts happen from time to time and we do not want to scare people with failing PR checks

.github/workflows/automerge.yml
abxhr: I thought to only perform `autoupdate` to the PRs with that label added. 
I'll get this removed.

.github/workflows/automerge.yml
abxhr: I thought to maybe have `autoupdate` on the pull request branch, each time a commit is made onto `master`. So I guess this might be needed, upto what I've seen from [this example repo](https://github.com/chinthakagodawita/autoupdate-test/blob/main/.github/workflows/autoupdate.yml), where `autoupdate` was implemented.

Let me know otherwise.

derberg: hmmm.... but the example that you linked works on a repo where PRs are created from branch on an upstream, this is why workflow is configured there to react on any branch commit.

with AsyncAPI, every "human" PR comes from a fork, every "bot" PR comes from upstream branch

so we need to make a decision here: do we really want to support both flows or the one that is critical here - "bot" one

to keep things simple:
- I think we need to focus on "bot" only. 
- autoupdate cannot be part of `automerge.yml` file but be independent and run the same way is it is done with linked example, on every push to any branch

makes sense?

abxhr: Yeah, I think that makes sense now. (*Although I'm a bit confused with the fork and upstream branch part*)

I'll make an independent file (`autoupdate.yml`), and remove this workflow from `automerge.yml`. As we need to focus on "bot" only, I'll change the `push` to any branch instead of just `master`.

derberg: @abxhr let us have a look at `.github` repo as an example.

`asyncapi/.github` is an upstream. When you work on this PR, you do not create a branch in the upstream repo, you first create a fork `abxhr/.github` and then a branch in the fork. This is a "human" flow in AsyncAPI. Even maintainers like me work the same way, fork flows only.

Now, in case of AsyncAPI bot user, or dependabot from GitHub - they do not create forks. They are admins in organization, and create branches in upstream, in for example `asyncapi/.github` and not `asyncapi-bot/.github`

So, the workflow that you will create, that reacts only on push to any branch, will never run for "humans" because when you push to a branch in `abxhr/.github`, the workflow from `asyncapi-bot/.github` knows nothing about it. Therefore `autoupdate` will only work for "bot" workflows.

2 things that make me curious when I look on the example that you linked:
- what if the commit is done to default branch, main/master, what will happen?
- if I have a branch `test` and do not have a PR opened, I guess autoupdate still runs (would be super awesome as it would solve one other issue we have in the release process)

abxhr: OH! It all actually makes sense to me now 😅
- If a commit is done to `main`/`master`, I guess that commit would also be updated on that PR branch (in `upstream`).
- That would be pretty cool! But would it be possible using this workflow template?

abxhr:
@derberg Thank you for being so helpful and patient! 

I have made the requested changes ✅

derberg (approved):
Awesome work! Thanks for you patience in the review!

derberg:
it doesn't work as we planned https://github.com/asyncapi/.github/issues/99 😅

abxhr:
@derberg Oh no! 😔
I guess we could fix it by replacing the labeler job, with an action that just labels the PR, as you mentioned in the [issue](#99)

derberg:
@abxhr yeah, nothing to worry about, these things happen, we just need to improve. You can't forsee everything

PR #86, opened 2021-10-05
ci: update stale workflow to ignore specific labels

derberg:
unfortunately, I did not notice that new version of the Stale action changed a bit the config property name by adding plural `s` 🤦🏼 

as a result all issues and PRs with `keep-open` were marked by stale action as candidates for closure anyway.

sorry, I will review all the stale bot emails from today and do some cleanup of those at the same time


magicmatatjahu (approved):
LGTM! :)

PR #103, opened 2021-11-25
ci: ignore `infra` repo

smoya:
**Description**

https://github.com/asyncapi/infra don't need to run workflows from org repository. No need for releasing, etc for now. In the future if needed we could add some automations


derberg:
just keep in mind there will be for example no workflow that notify us in slack about new issues and PRs

smoya:
> just keep in mind there will be for example no workflow that notify us in slack about new issues and PRs

Good point. Ok, I think we can close this one. Most of the workflows won't run as they won't detect the language so 👍

derberg:
yes, I think so yes

PR #106, opened 2021-12-08
ci: improve removal of ready-to-merge label only when present

derberg:
the problem with the current workflow that secures merging from "evil" contributors is that it doesn't check if the `ready-to-merge` label is even there, so it basically fails as it tries to remove a label that is not added to PR

### DIFF
--- a/.github/workflows/automerge-for-humans-remove-ready-to-merge-label-on-edit.yml
+++ b/.github/workflows/automerge-for-humans-remove-ready-to-merge-label-on-edit.yml
@@ -20,9 +20,16 @@ jobs:
         with:
           github-token: ${{ secrets.GH_TOKEN }}
           script: |
+            const labelToRemove = 'ready-to-merge';
+            const labels = context.payload.pull_request.labels;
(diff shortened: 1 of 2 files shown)


.github/workflows/automerge-for-humans-remove-ready-to-merge-label-on-edit.yml
```diff
@@ -20,9 +20,16 @@ jobs:
...
+            
             github.rest.issues.removeLabel({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              name: 'ready-to-merge'
-            }) 
\ No newline at end of file
+              name: labelToRemove
+            })
```
magicmatatjahu: Could you explain why you check label on PR, but remove on issue? Or maybe for Github rest `issueNumber` === `PRNumber`? 🤔

magicmatatjahu (dismissed):

magicmatatjahu (approved):

PR #107, opened 2021-12-08
fix: typo in help message

fmvilas:
Just fixing what I think is a typo.

PS: first PR from in-browser VS Code 🎉  Next time gotta remember to work on a branch instead of master 😂


derberg (approved):
LGTM, but I will just merge late afternoon as it is not critical and problem is that it has to be replicated everywhere and it is causing CI each time to 🐢

fmvilas:
No problem 👍

PR #114, opened 2022-01-03
ci: include pr number in commit message when automerging

derberg:
As @jonaslagoni noticed [here](https://github.com/asyncapi/modelina/pull/547#issuecomment-1004064050) automerge that I recently introduced do not include the number of the PR that was behind the squashed commit. I fix it with this PR and also fix it for automerging for bots too

I tested it [here](https://github.com/derberg/test-experiment/blob/master/.github/workflows/automerge-for-humans.yml) and result was [this](https://github.com/derberg/test-experiment/commit/7196d94002c445d96d9e416c23acfeef800b5ec0)


jonaslagoni (approved):
LGTM, thanks for the fast fix 🙏

KhudaDad414 (approved):
🚀

PR #117, opened 2022-01-05
ci: generate docker image in nodejs-release

magicmatatjahu:
**Description**

Generate docker image in `nodejs-release`:
- add needed envs to push Docker image to DockerHub

If someone want to release docker image have to install `@semantic-release-plus/docker` package in the project and add needed config to the `semantic-release` config. Also need to define Docker Image's build script in the `release` script. Example PR: https://github.com/asyncapi/server-api/pull/24

**Related issue(s)**
Part of https://github.com/asyncapi/server-api/pull/24

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -68,7 +68,7 @@ jobs:
         name: Install dependencies
         run: npm install
       - if: steps.packagejson.outputs.exists == 'true'
-        name: Release to any of NPM, Github, and Docker Hub
+        name: Publish to any of NPM, Github, and Docker Hub
         id: release
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -47,7 +47,7 @@ jobs:
 
   release:
     needs: test
-    name: Publish to NPM and GitHub
+    name: Publish to NPM/DockerHub and GitHub
     runs-on: ubuntu-latest
     steps:
       - name: Set git to use LF #to once and for all finish neverending fight between Unix and Windows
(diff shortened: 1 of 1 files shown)


.github/workflows/if-nodejs-release.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
   release:
     needs: test
-    name: Publish to NPM and GitHub
+    name: Publish to NPM/DockerHub and GitHub
```
smoya, line 50: Just small suggestion that makes sense to me (but can be obviously discarded)
```suggestion
    name: Publish to any of NPM, Github, and Docker Hub
```

.github/workflows/if-nodejs-release.yml
```diff
@@ -68,11 +68,16 @@ jobs:
         name: Install dependencies
         run: npm install
       - if: steps.packagejson.outputs.exists == 'true'
-        name: Release to NPM and GitHub
+        name: Build Docker Image (if script exists)
```
smoya, line 71: I would just delegate this to the `npm run release` script.

magicmatatjahu: Good idea! I will change it :)

smoya (dismissed):
LGTM 👍

.github/workflows/if-nodejs-release.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         name: Install dependencies
         run: npm install
       - if: steps.packagejson.outputs.exists == 'true'
-        name: Publish to any of NPM, Github, and Docker Hub
+        name: Release to any of NPM, Github, and Docker Hub
```
smoya, line 71: I would keep `Publish` word here for consistency.

magicmatatjahu:
@derberg You should be interested in this solution :) Of course it is not obligatory, maybe you have a better idea.

derberg (approved):
@magicmatatjahu definitely something we can add, release pipeline is still super flexible! especially in cases like generator, where we build docker in another workflow to build it from npm release and not sources. Nicely done!

derberg:
let's merge it later in the day or I can if you want tomorrow morning. These updates always occupy CI a bit during the day

Issue #116, opened 2022-01-05
Bumping versions often stalls release pipeline


jonaslagoni:
#### Describe the bug
When the node.js release pipeline bumps the package version often times GH stalls the CI runs, meaning the release pipeline is completely stuck.  See https://github.com/asyncapi/modelina/pull/561

Because of the changes introduced in the version bump `chore(release): ...` we dont need to run all the workflows and in theory skip them.

derberg:
damn, forgot about one thing, `skip ci` skips testing, but it is required on repo level...so these checks will be blocking merge 🤦🏼

derberg:
some info here https://github.community/t/github-pull-request-is-waiting-for-required-status-checks-even-with-ci-skip-in-commit-message/827

I added a comment as well

KhudaDad414:
Why don't we trust the bot and skip testing in case the author is `asyncapi-bot` [here](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L12)? any side effects?

derberg:
@KhudaDad414 oh man 🤯 I honestly did not take it into account 🤦🏼 
I was always declining it as `asyncapi-bot` cannot always be trusted, because sometimes `asyncapi-bot` sends PRs that are bump of package version, so tests must run to make sure that for example a bump of parser in generator will not break generator.

But we can also figure it out from the PR title I guess, just like we do it [here](https://github.com/asyncapi/generator/blob/master/.github/workflows/update-docs-on-docs-commits.yml#L26) 🤔 

so we could always trust allcontributors bot and asyncapi-bot in certain situations...

derberg:
@KhudaDad414 it didn't work. In all 59 repos (if we had branch protection set properly everywhere) the update of worklows failed because the tests were not triggered, so branch protection was not satisfied. Tests were not triggered because the test workflow was properly skipped because of new condition. So your idea worked but partially because workflow is skipped, jobs are not triggered, so PRs hang like here https://github.com/asyncapi/modelina/pull/614

Can you try closing these PRs and make a followup PR and move `if` form job to steps. It is not as grand as initial idea but still improvement that we need imho

KhudaDad414:
@derberg I don't have enough permission to close the PRs.

derberg:
@KhudaDad414 no worries. I'll close them

KhudaDad414:
Left a comment and it was already merged in the following repos:
```
dotnet-nats-template
[...]

PR #121, opened 2022-01-12
fix: include all semantic-release default branches for the release workflow + fix Genie

smoya:
**Description**

This PR adds all the default semantic-release branches to the release workflow for node apps.
See [semantic-release docs](https://semantic-release.gitbook.io/semantic-release/usage/configuration#branches).

This PR will also make Genie behave properly, who has been offering help randomly to people that didn't ask for it 😅 .

**Related issue(s)**
Relates to https://github.com/asyncapi/parser-js/pull/442#discussion_r783098429

### DIFF
--- a/.github/workflows/help-command.yml
+++ b/.github/workflows/help-command.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   create_help_comment_pr:
-    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') 
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-create-comment@v1
@@ -26,7 +26,7 @@ jobs:
             - `/ready-to-merge` or `/rtm` - This comment will trigger automerge of PR in case all required checks are green, approvals in place and do-not-merge label is not added
             - `/do-not-merge` or `/dnm` - This comment will block automerging even if all conditions are met and ready-to-merge label is added
   create_help_comment_issue:
-    if: !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'
+    if: ${{ !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-create-comment@v1
--- a/.github/workflows/help-command.yml
+++ b/.github/workflows/help-command.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   create_help_comment_pr:
-    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') 
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-create-comment@v1
@@ -26,7 +26,7 @@ jobs:
             - `/ready-to-merge` or `/rtm` - This comment will trigger automerge of PR in case all required checks are green, approvals in place and do-not-merge label is not added
             - `/do-not-merge` or `/dnm` - This comment will block automerging even if all conditions are met and ready-to-merge label is added
   create_help_comment_issue:
-    if: ${{ !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'}}
+    if: !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-create-comment@v1
--- a/.github/workflows/help-command.yml
+++ b/.github/workflows/help-command.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   create_help_comment_pr:
-    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') 
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-create-comment@v1
--- a/.github/workflows/help-command.yml
+++ b/.github/workflows/help-command.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   create_help_comment_pr:
-    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') }}
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/if-nodejs-release.yml
```diff
@@ -10,7 +10,10 @@ on:
       # below lines are not enough to have release supported for these branches
       # make sure configuration of `semantic-release` package mentiones these branches
       - next
-      - '**-release'
+      - next-major
+      - beta
+      - alpha
+      - '**-release' # custom
```
smoya, line 16: I didn't find this one on the semantic-release docs.

.github/workflows/if-nodejs-release.yml
```diff
@@ -10,7 +10,10 @@ on:
       # below lines are not enough to have release supported for these branches
       # make sure configuration of `semantic-release` package mentiones these branches
       - next
-      - '**-release'
```
smoya, line 13: The docs also show `'+([0-9])?(.{+([0-9]),x}).x` as regex for the branch name.I think we should also add it but waiting for feedback.

See https://semantic-release.gitbook.io/semantic-release/usage/configuration#branches

.github/workflows/if-nodejs-release.yml
derberg: this one is custom, driven by the discussion we had around release cycle and branch naming for specification where the branch name contains the release month.

.github/workflows/if-nodejs-release.yml
derberg: did you check if GH Actions support such regex, I don't think so afaik

smoya: No it doesn't. It allows basic filtering but can't formulate something close to that regex. I would just skip this regex for now.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

smoya:
> lgtm,
> 
> please make additional bug fix not related to your change but just doesn't make sense to overload CI with tiny fixes, so we can add it here:
> 
> * https://github.com/asyncapi/.github/blob/master/.github/workflows/help-command.yml#L13 change from `github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') ` to `github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot' `
> * https://github.com/asyncapi/.github/blob/master/.github/workflows/help-command.yml#L29 change from `${{ !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'}}` to `!github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'`
> 
> 🙏🏼

Ok, time to sound picky here 😅 :
I don't think we should add unexpected changes to this PR. The reason is that the merged commit name will provide insights about what it is doing, and in case something goes wrong when merging, spotting the bug will become harder as some side effects where added. People won't know what really the changes were by reading the commit title and this is, IMHO, a bad practice.
I think we also want other people to don't mix changes on the same PR unless they are 100% related and described.

Therefore, I created a new PR for those changes you suggested here: https://github.com/asyncapi/.github/pull/124

smoya:
> > lgtm,
> > please make additional bug fix not related to your change but just doesn't make sense to overload CI with tiny fixes, so we can add it here:
> > 
> > * https://github.com/asyncapi/.github/blob/master/.github/workflows/help-command.yml#L13 change from `github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') ` to `github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot' `
> > * https://github.com/asyncapi/.github/blob/master/.github/workflows/help-command.yml#L29 change from `${{ !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'}}` to `!github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'`
> > 
> > 🙏🏼
> 
> Ok, time to sound picky here 😅 : I don't think we should add unexpected changes to this PR. The reason is that the merged commit name will provide insights about what it is doing, and in case something goes wrong when merging, spotting the bug will become harder as some side effects where added. People won't know what really the changes were by reading the commit title and this is, IMHO, a bad practice. I think we also want other people to don't mix changes on the same PR unless they are 100% related and described.
> 
> Therefore, I created a new PR for those changes you suggested here: #124

As @derberg requested, I added the changes in this PR.

derberg:
I also do not like mixing, just here is exception, not code but ci. We stop doing it once we get https://github.com/asyncapi/.github/issues/116 solved, so basically have solution to skip ci entirely in such PRs

derberg (commented):
LGTM 🚀 

@KhudaDad414 anything from your side?

.github/workflows/help-command.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             - `/ready-to-merge` or `/rtm` - This comment will trigger automerge of PR in case all required checks are green, approvals in place and do-not-merge label is not added
             - `/do-not-merge` or `/dnm` - This comment will block automerging even if all conditions are met and ready-to-merge label is added
   create_help_comment_issue:
-    if: ${{ !github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot'}}
```
KhudaDad414, line 29: @derberg are you sure about this one? It gives syntax error without the brackets, at least in my fork. https://github.com/KhudaDad414/.github/actions/runs/1717756142

KhudaDad414 (changes requested):
LGTM.

.github/workflows/help-command.yml
smoya: Yes, it does. Missing tag. I thought this was handled by github when running the workflow. Thoughts @derberg?

derberg: I'm pretty sure I read in docs it is fine, but have nothing against adding them, just to be save, but then add it to all ifs in this workflow please

smoya: I added back the syntax expression format `{{`` and `}}` on the `create_help_comment_issue`.

.github/workflows/if-nodejs-release.yml
KhudaDad414: @smoya thanks! can you change https://github.com/asyncapi/.github/blob/41d14e04e38d75a6123f898d894e36e42dfde72a/.github/workflows/help-command.yml#L13 to `${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help' ) && github.actor != 'asyncapi-bot' }}` for consistency?

smoya: done!

.github/workflows/help-command.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   create_help_comment_pr:
-    if: github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') 
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help' && github.actor != 'asyncapi-bot') }}
```
KhudaDad414, line 13:
```suggestion
    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help') && github.actor != 'asyncapi-bot' }}
```

KhudaDad414 (approved):

KhudaDad414 (approved):
LGTM!

smoya:
Thanks folks for the review!

Anything else to add @derberg ? 🚀

derberg (approved):

PR #131, opened 2022-02-07
ci: new workflow to notify TSC members in slack

derberg:
This workflow throws a message in a dedicated channel when there is a mention of the GitHub team that groups all technical steering committee members:
- new discussion, issue or PR that mentions the team
- new or edited comment in issue or PR  that mentions the team (message links back to the comment)
- new comment in discussion and new comment/reply to a comment in discussion (message links back to the comment)
- reopened issue or PR that mentions the team

Because of `if` condition in the job, and not steps, workflow is skipped if TSC members are not mentioned

I tested this workflow in new channel `95_bot-tsc-members-mentioned`. Below you can see example messages that were triggered from my [test repo](https://github.com/derberg/test-experiment/actions/runs/1807131912).

Screenshot 2022-02-07 at 16 07 26

See also https://github.com/asyncapi/community/issues/253
This PR provides notifications to slack only for mentions.

Once we have a voting process facilitated, and we know how to do it, we will need to extend this workflow with new job that triggers a slightly different message, `@channel` with call for voting

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -15,6 +15,7 @@ on:
 
 jobs:
   bump:
+    if: startsWith(github.event.commits[0].message, 'chore(release):')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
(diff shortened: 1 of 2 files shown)


KhudaDad414 (approved):
LGTM!

derberg:
@KhudaDad414 thanks! I'll merge it later today as the other workflow update is in progress and it will take probably an hour to replicate

PR #142, opened 2022-03-10
ci: add lighthouse ci for tracking website performance

HarshCasper:
**Description**

- Add a `lighthouse-ci-test.yml` to track web performance 
- Add a `lighthouserc.json` to define the Lighthouse configuration.

**Related issue(s)**

See also https://github.com/asyncapi/website/issues/89

### DIFF
--- a/.github/workflows/lighthouse-ci-test.yml
+++ b/.github/workflows/lighthouse-ci-test.yml
@@ -1,0 +1,69 @@
+name: Lighthouse CI
+
+on:
+  pull_request:
+    branches:
+      - master
+    types: [opened, reopened, synchronize, ready_for_review]
+
+jobs:
+  lighthouse-ci:
+    name: Lighthouse CI
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


KhudaDad414:
Hey @HarshCasper, thanks for the contribution. I think this workflow belongs to the website repo. Is there a reason why we should have it here instead on the `website` repo itself?

HarshCasper:
Hi @KhudaDad414 — Thanks for your comment! Quoting @magicmatatjahu from this issue (https://github.com/asyncapi/server-api/issues/50):

> However, testing is done by this workflow https://github.com/asyncapi/server-api/blob/master/.github/workflows/if-nodejs-pr-testing.yml but we cannot update it, because the source for this workflow is located here https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-pr-testing.yml Even if you change something on this repo (in the source of that workflow) in future it will be overwritten by asyncapi-bot.

I suppose it would stand true here as well. All of the workflows specific to every project under the AsyncAPI is hosted here. It would make sense if we have the LightHouse CI workflow hosted here as well. I would love to get @derberg thoughts as well on this!

derberg (commented):
this one is ❤️ 

before we make review please move it to `website` repo. This workflow is super custom, only for the website so should not land in `.github` repo. Whatever lands here in `.github` is then replicated to all other repositories, and we do not want it in this case.

PR #143, opened 2022-03-10
feat: updated org profile README banner

mcturco:
[019bfa09bc924b51a3bc04ba0e20e1f5-with-play](https://www.loom.com/share/019bfa09bc924b51a3bc04ba0e20e1f5)

**Description**
I am proposing to update the org profile README banner from the Eve & Chan animation with the old logo to a banner that represents the new brand.

In the future, when Eve & Chan gets reimagined within the new brand, I think we can bring them back here. But for now I think it makes more sense to switch it to something that matches the new brand.


alequetzalli (approved):
🚢

Issue #115, opened 2022-01-03
PR owner can't see "branch out of date with master" notice after `/rtm`


mcturco:
#### Describe the bug
When adding `/rtm` on my approved PR, the expected behavior is for GH bot to automatically rebase the branch with the master version. Either way, even though the PR branch was out of date, the PR owner should be able to see a notice of that even when they aren't a codeowner. 

Here's the only notice that I saw:
![Screen Shot 2022-01-03 at 11 34 23 AM](https://user-images.githubusercontent.com/60163079/147958790-3738b74e-fe37-42c0-91fd-344a5e2bf2e1.png)

From @derberg:
"we need to think about fix as there are 2 ways to fix it, bot can do update with master or bot can just provide a comment feature like /rtm so anyone can trigger merge….not sure if the first solution will be accepted by all

we do autoupdates already but only in PRs created by bot, as nobody cares about bot’s feelings (this is why robots will kill all humans one day :smile: )

https://github.com/asyncapi/website/pull/526 notice that github bot updated branch with latest master"

derberg:
more context on why not being up to date with master blocks merge, we have such setting everywhere 👇🏼 

![Screenshot 2022-01-03 at 18 10 57](https://user-images.githubusercontent.com/6995927/147959188-8a9057fc-eb54-40e2-93a2-a618c3a8b9a0.png)

This is a best practice. One way to skip this setting and resolve this problem the other way around is that our testing pipelines should be designed in the way, that when they run, they first rebase with master (to testing always happens with latest changes from master), and they are simply triggered on any merge to master.

not sure what is the best solution really. I think we should pick the one that is least heavy for the CI (to not overload it) and doesn't matter how complicated it will be to achieve it.

derberg:
This is what people with write access to the repo see

![Screenshot 2022-01-03 at 17 13 37](https://user-images.githubusercontent.com/6995927/147959692-59f4df6b-b6cb-4598-9679-77768a4d8ad9.png)

KhudaDad414:
I think the first option, having a command like `/rebase` would be less heavy on CI and provide a way for non-code-owners to rebase. plus, adding a comment like `This branch is out-of-date with the base branch, you can use /rebase to update this branch.` from `/rtm` if it fails due to out-of-date branch maybe?

PR #146, opened 2022-03-21
ci: let autoupdate.yml run on pull_request

smoya:
**Description**

I’m thinking that the `autoupdate.yml` action that keeps release branches up-to-date with the destination branch (e.g. master ) should also run on creating a PR and not only on push event.
https://github.com/asyncapi/spec/blob/master/.github/workflows/autoupdate.yml#L13-L15

I recently created few PR’s in order to keep the release branches (`2022-04-release`) up-to-date with `master` branch . But I don’t want to wait until someone else pushes changes on those release branches. Instead, I would like the bot to update them ASAP. And this is not possible since the action is triggered only by the push event.

Does it make sense to also trigger it on `pull_request` event?
If not, please discard this PR.

### DIFF
--- a/.github/workflows/autoupdate.yml
+++ b/.github/workflows/autoupdate.yml
@@ -17,6 +17,13 @@ on:
       - 'dependabot/**'
       - 'bot/**'
       - 'all-contributors/**'
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+    branches-ignore:  
+      - 'version-bump/**'
+      - 'dependabot/**'
+      - 'bot/**'
+      - 'all-contributors/**'
 
 jobs:
   autoupdate-for-bot:


derberg:
@smoya I'm concerned that this workflow will run on any PR event to master branch. So a lot of triggers with no effect.

maybe better we should come up with some command like `/update` that would call proper workflow that updates the base branch of PR where `/update` comment was made?

somehow related to https://github.com/asyncapi/.github/issues/115

smoya:
We could also let the action to be triggered under demand with the `workflow_dispatch` event.

derberg:
@smoya just keep in mind only codeowners, folks with maintainers rights can trigger such workflow event

smoya:
> @smoya just keep in mind only codeowners, folks with maintainers rights can trigger such workflow event

yikes! Yeah, that is not ideal :(

KhudaDad414:
@derberg If I understand it correctly, we need to do the `/update` command in cascade. like if we have `master -> 2022-04-release -> some-patch`, first we have to update `2022-04-release` with `master`, then we can update `some-patch` with `2022-04-release`. 
Getting changes from `master` to `2022-04-release` is a little bit tricky.

derberg:
@KhudaDad414 sorry, I didn't get it. Why `Getting changes from master to 2022-04-release is a little bit tricky.` ?
This is what `autoupdate-action` is going. Basically merging master into the base branch, as long as there are no conflicts of course.

Maybe we can even use the same action, but triggered not by push but by PR comment with `/update`, no?

derberg:
if it is label based then it means it will work only once, right? and it won't be much useful for #115 ? cause sometimes you might need to request update more than once

KhudaDad414:
> if it is label based then it means it will work only once, right?

Not necessarily in this case. since `autoupdate.yml` will run on `push` event as well, having `auto-update` label is the condition for being updated. In theory, I can call `/auto-update` on any PR and the PR's branch will be kept updated till it is merged.

.github/workflows/autoupdate.yml
```diff
@@ -17,6 +17,13 @@ on:
       - 'dependabot/**'
       - 'bot/**'
       - 'all-contributors/**'
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
```
alequetzalli, line 21: What is the `synchronize` PR type for? 🤔

alequetzalli:
> * create a new label called `auto-update` for PRs.
> * create a new command called `/auto-update` or `/update`

I like @smoya's  idea and what @KhudaDad414 suggests here:

- create a new label called `auto-update` for PRs.
- create a new command called `/auto-update` or `/update`

derberg:
closing in favour of https://github.com/asyncapi/.github/pull/149

PR #147, opened 2022-03-22
ci: add a workflow to build docker images on pull requests

HarshCasper:
**Description**

- Add a workflow to build Docker images on PRs
- Implement CI caching using the Cache backend API

**Related issue(s)**

https://github.com/asyncapi/server-api/issues/50

### DIFF
--- a/.github/workflows/global-replicator.yml
+++ b/.github/workflows/global-replicator.yml
@@ -96,3 +96,19 @@ jobs:
           committer_email: info@asyncapi.io
           commit_message: "ci: update generic workflows"
           repos_to_ignore: shape-up-process,glee-hello-world
+
+  replicate_docker_workflows:
+    name: Replicate Docker workflows needed for any project
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Replicating file
+      uses: derberg/copy-files-to-other-repositories@v1.0.0
+      with:
+        github_token: ${{ secrets.GH_TOKEN }}
+        topics_to_include: docker
+        patterns_to_include: .github/workflows/if-docker-pr-testing.yml
+        committer_username: asyncapi-bot
+        committer_email: info@asyncapi.io
+        commit_message: "ci: update workflows for docker-based projects"
--- a/.github/workflows/if-docker-pr-testing.yml
+++ b/.github/workflows/if-docker-pr-testing.yml
@@ -1,6 +1,6 @@
 #This action is centrally managed in https://github.com/asyncapi/.github/
 #Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
-#It does magic only if there is package.json file in the root of the project
+#It does magic only if there is a Dockerfile in the root of the project
 name: PR testing - if Docker
 
 on:
@@ -38,13 +38,13 @@ jobs:
       - if: steps.docker.outputs.exists == 'true'
         name: Extract metadata for Docker
         id: meta
-        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        uses: docker/metadata-action@v3
         with:
           images: ${{ env.IMAGE_NAME }}
 
       - if: steps.docker.outputs.exists == 'true'
-        name: Build and push Docker image
-        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        name: Build Docker image
+        uses: docker/build-push-action@v2
         with:
           context: .
           push: false
--- a/.github/workflows/if-docker-pr-testing.yml
+++ b/.github/workflows/if-docker-pr-testing.yml
@@ -1,0 +1,54 @@
+#This action is centrally managed in https://github.com/asyncapi/.github/
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+#It does magic only if there is package.json file in the root of the project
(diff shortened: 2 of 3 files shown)


HarshCasper:
cc: @smoya @derberg

.github/workflows/if-docker-pr-testing.yml
```diff
@@ -0,0 +1,54 @@
...
+        shell: bash
+
+      - if: steps.docker.outputs.exists == 'true'
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - if: steps.docker.outputs.exists == 'true'
+        name: Extract metadata for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
```
derberg, line 41: why not using `v3` but specific commit?

.github/workflows/if-docker-pr-testing.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          images: ${{ env.IMAGE_NAME }}
+
+      - if: steps.docker.outputs.exists == 'true'
+        name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
```
derberg, line 47: why not using `v1` but specific commit?

derberg (changes requested):
looks good 🚀 
just left few comments

alequetzalli:
This sounds amazing, @HarshCasper!

I approve it, but obviously looks like @derberg and other folks will have more feedback on the code to make sure it works 🎉

alequetzalli (approved):

.github/workflows/if-docker-pr-testing.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - if: steps.docker.outputs.exists == 'true'
+        name: Extract metadata for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.IMAGE_NAME }}
+
+      - if: steps.docker.outputs.exists == 'true'
+        name: Build and push Docker image
```
magicmatatjahu, line 46: We don't push image.

```suggestion
        name: Build Docker image
```

derberg:
@HarshCasper yo, do you plan to continue working on this?

.github/workflows/if-docker-pr-testing.yml
```diff
@@ -0,0 +1,54 @@
...
+#Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+#It does magic only if there is package.json file in the root of the project
+name: PR testing - if Docker
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+env:
+  IMAGE_NAME: ${{ github.repository }}
```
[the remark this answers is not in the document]
HarshCasper: Yes, it would. Workflow-level `env` is supported. Please refer: https://docs.github.com/en/actions/learn-github-actions/environment-variables#:~:text=The%20entire%20workflow%2C%20by%20using%20env%20at%20the%20top%20level%20of%20the%20workflow%20file.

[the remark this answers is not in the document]
derberg: oh cool!

different docs, different answers

cc @KhudaDad414 look, I think some time ago we wanted to use it this way but we only found docs that said that only on job level it is possible

derberg:
@magicmatatjahu @smoya do we want it only on server-api for now? or `generator` too?

@HarshCasper please add new step to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml that is replicating this new workflow only to repositories with `docker` topic

magicmatatjahu:
@derberg Atm we can "test" workflow only in the `server-api` repo, but long term we should run it on every repo where we publish docker image.

.github/workflows/if-docker-pr-testing.yml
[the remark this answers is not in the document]
KhudaDad414: I guess since the variable is used in `with:` part which ends up in the runner itself it should be good. but in the previous case, I tried to use it in the workflow syntax itself (in `if:` part) which runs outside of the runner (by GitHub Actions somewhere) and it wasn't working 🤔

derberg:
ok then, just added `docker` topic to `server-api`

HarshCasper:
Done @KhudaDad414!

derberg:
@HarshCasper fyi below left todo to get this merged

> @HarshCasper please add new step to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml that is replicating this new workflow only to repositories with docker topic

.github/workflows/global-replicator.yml
```diff
@@ -95,4 +95,19 @@ jobs:
...
+  replicate_docker_workflows:
+    name: Replicate Docker workflows needed for any project
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Replicating file
+      uses: derberg/copy-files-to-other-repositories@v1.0.0
+      with:
+        github_token: ${{ secrets.GH_TOKEN }}
```
derberg, line 109:
```suggestion
        github_token: ${{ secrets.GH_TOKEN }}
        topics_to_include: docker
```

we want this workflow to be replicated only to repos that have `docker` topic, which is `server-api` repo so far

Issue #92, opened 2021-10-20
Readme links check


panwauu:
I found some broken links in .md files and googled if there are tools to report broken tools and found this [github action](https://github.com/gaurav-nelson/github-action-markdown-link-check) and this other [github action](https://github.com/ocular-d/md-linkcheck-action).
It would be nice to have this in all of the repos

derberg:
we should definitely use https://github.com/gaurav-nelson/github-action-markdown-link-check and have 2 workflows:
- 1 to check only modified files on a PR level
- 1 to periodically, cron job, like once a week check all files in a repo, and push notifications in case of failure to our slack, maybe docs channel @alequetzalli ?

alequetzalli:
hey @derberg, that sounds like a great idea and then getting a push notification in the `#docs` Slack channel is incredibly useful!

derberg:
whoever picks up this issue, please remember that workflow must be created in `.github` repo but initially please test it on your own test repo and then provide links in PR so we can check how it worked.

you do not have to test Slack integration, just adapt [the following steps](hub/blob/master/.github/workflows/notify-tsc-members-mention.yml#L41-L52) that we already use for other slack-notification-related workflows. We will provide you with proper webhook secret for `#docs` channel when needed

good luck ☮️

Namyalg:
I would like to work on this issue

derberg:
Please go ahead. Lemme know if you need help. Are you part of OpenForce?

Namyalg:
Yep

aniketh-varma:
Hi, would like to work on this issue. I am a part of OpenForce 2022.

derberg:
@aniketh-varma Hey, thanks but since @Namyalg was here first, the issue belongs to her. Please have a look at other issues from the list.

Namyalg:
As mentioned [here](https://github.com/gaurav-nelson/github-action-markdown-link-check), is there anything specific that should go into the `mlc_config.json` file ?

derberg:
@Namyalg hey, we are having some issues with automation:

- https://github.com/asyncapi/.github/blob/master/.github/workflows/link-check-cron.yml#L31 is a wrong env key, it should be `SLACK_WEBHOOK_URL: ${{ secrets.SLACK_DOCS_CHANNEL }}`
- after looking at https://github.com/asyncapi/spec/runs/5802713025?check_suite_focus=true#step:4:23 sounds like config file is required?

Namyalg:
[...]

PR #151, opened 2022-04-04
fix: rename env from  `SLACK_DOCS_URL` to `SLACK_WEBHOOK_URL`

Namyalg:
**Description**
Rename env from  `SLACK_DOCS_URL` to `SLACK_WEBHOOK_URL` in `link-check-cron.yml`

**Related issue(s)**

### DIFF
--- a/.github/workflows/link-check-cron.yml
+++ b/.github/workflows/link-check-cron.yml
@@ -28,5 +28,5 @@ jobs:
           status: ${{ job.status }}
           fields: repo,message,action,eventName,ref,workflow
       env:
-          SLACK_DOCS_CHANNEL: ${{ secrets.SLACK_DOCS_CHANNEL }}
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_DOCS_CHANNEL }}
       if: failure() # Only, on failure, send a message on the Slack Docs Channel (if there are broken links)
--- a/.github/workflows/link-check-cron.yml
+++ b/.github/workflows/link-check-cron.yml
@@ -26,7 +26,7 @@ jobs:
         # More information can be found here: https://github.com/tcort/markdown-link-check#config-file-format
         # The default file is mlc_config.json, located at the root of the directory
         
-        # config-file: name-of-config-file
+        # config-file: name-of-config-file.json
 
     - name: Report workflow run status to Slack
       uses: 8398a7/action-slack@v3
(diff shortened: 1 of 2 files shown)


derberg:
@KhudaDad414 what is your take on https://github.com/asyncapi/.github/issues/92#issuecomment-1087389838

I'm personally not 100% sure if we should maintain global config just like global workflow or maybe rather the config for links validator should be custom per repo? 🤷🏼

KhudaDad414:
@derberg IMO having the config file per repo seems like a better choice. At first, it is a little more work to be done for the links that are already a part of the codebase. 
But in the future, the dead links can be easily added to the config file using the same PR. the contributor doesn't have to open a separate PR in this repo to ignore the link.

Namyalg:
So based on the file being checked, a separate config file has to be used?
Is this correct ?

KhudaDad414:
@Namyalg  I don't think we need to do anything. for repositories that we need some custom configuration, we can just create a `mlc_config.json` file and it will be picked up by the action automatically(I guess).

derberg:
from the logs, it looks like no additional workflow config is needed and GH action always looks up for `mlc_config.json` in the default location. My understanding is that default location is `root` of the repo? @Namyalg can you please confirm?

Namyalg:
Yep, that's right

derberg:
@Namyalg we do not have a place where to put it, just put a comment in the workflow file, with link to config example please

Namyalg:
Alright will do

Namyalg:
Comments about the config file have been added

.github/workflows/link-check-cron.yml
```diff
@@ -21,12 +21,18 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'yes'
+        
+        # A configuration file can be included, indicating the properties of the link check action
+        # More information can be found here: https://github.com/tcort/markdown-link-check#config-file-format
+        # The default file is mlc_config.json, located at the root of the directory
+        
+        # config-file: name-of-config-file.json
```
derberg, line 29:
```suggestion
        # A configuration file can be included, indicating the properties of the link check action
        # More information can be found here: https://github.com/tcort/markdown-link-check#config-file-format
        # Create mlc_config.json file in the root of the directory
```

please update also the other comment in other workflows.
we should not encourage people to provide custom names for config files because then they would need to modify workflows, but they should not modify workflows as we will then overwrite their changes we make updates here in `.github` repo

Namyalg:
Comments are updated

derberg (approved):

alequetzalli (approved):

Issue #150, opened 2022-03-29
Validate workflow on PR level


KhudaDad414:
#### Reason/Context
Since we don't validate our workflow files at the PR level, there is no way for us to make sure that the workflow files are valid YAML files.

#### Description
We need a linter as a workflow to make sure our workflows are valid YAML files.
BONUS: if the workflow is able to validate against the [workflow schema](https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json) it would be much better.

Namyalg:
I am interested to work on this issue
My understanding of this issue is : 
- Create a workflow to check if new workflow files are valid and error-free using a linter

derberg:
exactly, additionally workflow should only run if changes are done to files under `.github/workflows` directory.

regarding validation against schema, have a look at https://github.com/asyncapi/community/blob/master/.github/workflows/validate-tsc-members.yml#L19 where I already did something similar using `Ajv`. You will just have to fetch schema from remote first, instead of storing it in the repo and reading as file

Namyalg:
Thank you for the reference, I'll get started on the implementation

Namyalg:
I plan to use the action https://github.com/tj-actions/changed-files, which will detect the addition or modification of a file.
Once the list files are obtained, the files with .yml are run against the schema validator.

derberg:
sounds good 👍🏼

Namyalg:
@KhudaDad414 , is this [workflow schema](https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json) the correct one ?
It doesnt seem to follow the exact format of the yml files written for GHA

KhudaDad414:
@Namyalg I am fairly sure that this is the schema.
Are there any issues when you validate workflows?
I tried a few of the workflows and they seem to validate fine.

Namyalg:
Yes, this is what I get when I validate the [linkcheck workflow](https://github.com/asyncapi/.github/blob/master/.github/workflows/link-check-cron.yml)
![image](https://user-images.githubusercontent.com/53875297/164689432-2e02aed8-368c-46c4-88de-874ad4cc745c.png)

KhudaDad414:
@Namyalg I suggest validating the file with the approach that @derberg mentioned here: https://github.com/asyncapi/.github/issues/150#issuecomment-1088445829

I tested it with this code and it works like a charm:
😃 
```js
[...]

PR #159, opened 2022-04-23
feat: validate workflow schema

Namyalg:
**Description**

This PR consists of a script and a workflow that validates the yml schema of a workflow file

**Related issue(s)**
Issue #150 

This is sample run on [2 erroneous files ](https://github.com/Namyalg/.github/runs/6137835602?check_suite_focus=true)


Namyalg:
I have created a workflow file and this is the log on [2 erroneous workflow files](https://github.com/Namyalg/.github/runs/6137960820?check_suite_focus=true)

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,35 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v18.7
+
+      - name: Install dependencies
+        run: |
+            npm install ajv@8.11.0
+            npm install axios
+            npm install js-yaml
+            
+      - name: Identify changed yml files # Identifies files added, modified, copied or renamed
```
derberg, line 30: please use `actions/github-script@v4`

you can just `require` your script inside like https://github.com/asyncapi/spec/blob/master/.github/workflows/new-spec-release.yml#L59-L61

derberg (changes requested):
Thanks a lot. I made few suggestions. Once you apply them, please show workflow in action in some test repo

.github/scripts/validate-workflow-schema.js
```diff
@@ -0,0 +1,36 @@
+function getFileExtension(filename){
+    return filename.split('.').pop();
+}
+
+async function validateYmlSchema(filename){
+    if(getFileExtension(filename) === 'yml'){
```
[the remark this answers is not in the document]
Namyalg: 👍

.github/scripts/validate-workflow-schema.js
```diff
@@ -0,0 +1,36 @@
...
+        const Ajv = require('ajv');
+        const axios = require('axios');
+        const yaml = require('js-yaml');
+        const fs = require('fs').promises;
+        console.log("File name " + filename);
+        const schema = await axios.get(
+        'https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json'
```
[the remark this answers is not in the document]
Namyalg: yep, that's better

.github/scripts/validate-workflow-schema.js
```diff
@@ -0,0 +1,36 @@
...
+            const valid = validator(target);
+            if (!valid) {
+                console.error(`Validation failed with the following errors:`);
+                console.log(validator.errors)
+            } else {
+                console.log("The workflow in " + filename + " follows the schema");
+            }
+        }
+        catch(err){
+            console.log("The workflow in " + filename + " has an invalid schema");
```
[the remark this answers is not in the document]
Namyalg: Thanks for this, I was looking out for something similar, as a workflow that passes even on failure may not be very intuitive to the user

.github/workflows/validate-workflow-schema.yml
Namyalg: Alright

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,35 @@
+name: Validate Workflow Schema
+
+on:
+  push:
```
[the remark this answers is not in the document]
Namyalg: 👍

.github/workflows/validate-workflow-schema.yml
Namyalg: @derberg, Is this a replacement for this ?https://github.com/Namyalg/.github/blob/eafbd93f29d67c9c03d824a5d23413455ea2cfc9/.github/workflows/validate-workflow-schema.yml#L32

derberg: yes, don't call code like this directly but through the `github-script` action

Namyalg:
@derberg , In case 2 yml files are created during a PR, and the first one fails, currently the action will stop at the first failed file. Is it a good idea to do this or validate both the files and then report the error ?

derberg:
@Namyalg best developer experience would be to validate all the files and throw error at the end of validation with list of all affected files, end related errors

just `catch` errors into a `Map` where you keep filename and error and then at the end, if `size` is bigger than `0` just throw error with all details

Namyalg:
Okay, working on this

Namyalg:
I have made the changes suggested

This is a [sample](https://github.com/Namyalg/.github/runs/6198894344?check_suite_focus=true) workflow run

Few points : 
- The JSON schema has been stored in a file in `scripts` . I was facing some issues due to the async call to axios, hence fs.readFileSync has been used
- On encountering errors in file(s), the name of the file and the log is shown. The workflow fails (using actions/@core)
- Without installing `actions/@core`, an package not found error occurred (As it was earlier mentioned that actions/@core is a part of the actions/github-script action)

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,36 @@
+name: Schema Validation
+
+on:
+  workflow_dispatch:
```
derberg, line 4: can be removed, we do not need to run it on demand, only on PR I think

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v18.7
+
+      - name: Install dependencies
+        run: |
+            npm install ajv@8.11.0
+            npm install axios
+            npm install js-yaml
+            npm install @actions/core
```
derberg, line 28: it is ok to leave it explicitly here, otherwise, you would have to pass it into your script like `script(files, core)`

but yeah, in case of this workflow where we anyway need to install some dependencies, it is fine

just huge request from my side: just like in case of `ajv@8.11.0` please also make sure to lock other dependencies to a specific version

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,36 @@
...
+    name: Validate workflow schema
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v18.7
```
derberg, line 21: can we make sure it check only changes in files in `.github/workflows` ?

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,36 @@
+name: Schema Validation
+
+on:
+  workflow_dispatch:
+  pull_request:
```
derberg, line 5: please narrow triggering only in case PR modifies something in the `.github/workflows` in `yml` or `yaml` files

ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,36 @@
+name: Schema Validation
```
derberg, line 1:
```suggestion
name: GitHub Workflows Schema Validation
```

derberg (changes requested):
looking better, left some comments, please have a look

.github/workflows/validate-workflow-schema.yml
Namyalg: Alright

Namyalg:
@derberg the changes have been made

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,43 @@
+name: GitHub Workflows Schema Validation
+
+on:
+  pull_request: 
+    paths:
+        - '**.yml'
```
derberg, line 6: can't we do `.github/workflows` there too?

Namyalg: Yes, the path is now updated to `.github/workflows/**.yml` and `.github/workflows/**.yaml`

derberg:
@KhudaDad414 anything from your side. I'm ok with the current shape, so from my point of view examples can be removed, but maybe you wanna have a look into logs too

derberg:
@KhudaDad414 hey, can you have a look?

.github/scripts/validate-workflow-schema.js
```diff
@@ -0,0 +1,57 @@
+// Dependencies
+const core = require('@actions/core');
+const Ajv = require('ajv');
+const yaml = require('js-yaml');
+const fs = require('fs');
+
+function validateYmlSchema(filename){
+    // Read the schema and workflow file synchronously
+    let schema = fs.readFileSync('.github/scripts/yml-schema.json', {encoding:'utf8', flag:'r'});
```
KhudaDad414, line 9: @Namyalg can't we use the [schema store](https://json.schemastore.org/github-workflow.json) hosted file instead of our own copy? I don't think it is advisable from a maintainability point of view since **schema store** is actively maintaining this file.

Namyalg: As in read directly from the URL and compare?

KhudaDad414: Yes, I have done some tests [here](https://github.com/KhudaDad414/.github/blob/bdb585cda6833a3a57ac7b7276892920238d6f07/.github/scripts/validate-workflow-schema.js) and it works like a charm.

derberg:
@Namyalg hey, do you need some help?

Namyalg:
Hey @derberg, I was wondering how I should go about changing my code
Do we want to include [this version](https://github.com/KhudaDad414/.github/blob/bdb585cda6833a3a57ac7b7276892920238d6f07/.github/scripts/validate-workflow-schema.js#L36) of the function or just replace the [first line](https://github.com/Namyalg/.github/blob/6c6498a11be74250619272d8f632941da4a27765/.github/scripts/validate-workflow-schema.js#L9) and read from the URL instead ?

derberg:
@Namyalg hey, best is just to remove `yml-schema.json` and read it from URL provided by @KhudaDad414

derberg:
@Namyalg I see that you are already installing `axios` as dependency, so you can use it

derberg:
@Namyalg hey, need a hand?

Namyalg:
Hey @derberg, working on the changes

derberg:
@Namyalg fyi there are still 2 not addressed comments

Namyalg:
Hey @derberg, apologize for the delay.
The script to validate schemas reads from the URL, based on the [reference script](https://github.com/KhudaDad414/.github/blob/bdb585cda6833a3a57ac7b7276892920238d6f07/.github/scripts/validate-workflow-schema.js) written by @KhudaDad414

derberg:
@Namyalg no worries 👍🏼 

@KhudaDad414 can you please have another look?

KhudaDad414 (dismissed):
Let's see this in action :) thanks @Namyalg 👏

Namyalg:
Shall I delete the test file [one](https://github.com/Namyalg/.github/blob/validate-workflow-schema/sample.yml) and [two](https://github.com/Namyalg/.github/blob/validate-workflow-schema/.github/workflows/sample.yml) ?

KhudaDad414:
@Namyalg sure, since we do not need them.

Namyalg:
Done with the deletions 👍

.github/workflows/validate-workflow-schema.yml
```diff
@@ -0,0 +1,43 @@
...
+          files: |
+            .github/workflows/*.yml
+            .github/workflows/*.yaml
+  
+      - name: Install dependencies
+        run: |
+            npm install ajv@8.11.0
+            npm install axios@0.27.2
+            npm install js-yaml@4.1.0
+            npm install @actions/core@1.8.0
```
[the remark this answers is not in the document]
KhudaDad414: hey @Namyalg, do you want to make this change? let's not keep this PR open for a year 😆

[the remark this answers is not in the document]
Namyalg: Yes, I will fix it asap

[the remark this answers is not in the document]
derberg: I don't think it has been addressed

derberg (changes requested):
We are almost ready to go, but since the creation of this PR, some changes were introduces in project in regards to "global" files.

- You need to add this new workflow explicitly to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L134 list
- You need to move JS script to [workflows/scripts](https://github.com/asyncapi/.github/tree/master/.github/workflows/scripts) and add there to the script a comment on what is this script used for. If for my other comment, you go with `package.json` and `package-lock.json` then you will need to add there a folder for the script and related files

Lemme know if you need more explanation

Namyalg:
Okay @derberg , will look into it

derberg:
@Namyalg need help?

derberg:
@Namyalg I hope you plan to complete it as you spent a lot of time on it.

We have one contributor that is willing to complete it for you, but I want to make sure you are ok with it

Namyalg:
Hello @derberg , working on it. 
I would like to collaborate with the other contributor.

derberg:
@14Richa I know you were interested to help with this one
@Namyalg thanks for quick response 🙏

14Richa:
Hey @Namyalg, let me know if you need any help. You can ping me on slack as well. :) 

Always happy to help!!

14Richa:
Thanks, @derberg.  I will get in touch with @Namyalg and make the pr ready to merge asap together. :)

derberg:
any success? maybe we should just start from scratch?

we need this validation as we have lots of workflows, and I caused some errors recently just because of syntax error in a workflow 🙃 😄

Namyalg:
Yes, I think we can start from scratch @derberg 
Where should I get started ?

derberg:
@Namyalg if you want to continue with this one, no need to start from scratch. There are minor stuff left. I suggested from scratch in case you want to let it go.

did you have a change to sync with @14Richa and get some help? Both of you are in our Slack

KhudaDad414:
closing since it has been done here: https://github.com/asyncapi/.github/pull/238

PR #163, opened 2022-05-10
ci: change token for autoupdate to use codeowner token

derberg:
Autoupdate do not work with forks -> https://github.com/asyncapi/website/runs/6361700980?check_suite_focus=true

Reason is that we use there token of wrong bot, `Eve` is the right one as `Eve` is a codeowner in all projects, so only `Eve` bot have `write` access to forks as a valid maintainer.

At least I think this is the solution 😆


KhudaDad414:
@derberg I don't see what the issue is here. and the run that you linked has been successful 🤔

KhudaDad414 (approved):

PR #170, opened 2022-06-22
ci: fix version bump workflow and make sure in future we get notified

derberg:
related to https://github.com/asyncapi/dotnet-nats-template/actions/runs/2536516647 and we learned that for at least a month the version bump workflow doesn't work

this PR:
- make sure bump version works the same way release and pr testing workflows
- add slack notifications in case of these workflows fail


jonaslagoni (dismissed):
LGTM 👍

derberg:
@fmvilas this one is pretty critical senior 😄

fmvilas (approved):

PR #174, opened 2022-07-22
ci: close staled issues with `not planned` reason

smoya:
**Description**

This PR forces the `stale-issues-prs` workflow to close staled issues with the `not planned` reason instead of `completed`, as per completed means a fix, feature, or any other work has been done in order to close it, which in our community, it doesn't happen; we manually close issues that has been fixed/resolved, and stale issues are normally things we won't do.  

**Related issue(s)**
Example of issue that has been marked as `completed` but shouldn't: https://github.com/asyncapi/spec/issues/596

### DIFF
--- a/.github/workflows/stale-issues-prs.yml
+++ b/.github/workflows/stale-issues-prs.yml
@@ -13,7 +13,7 @@ jobs:
     name: Mark issue or PR as stale
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/stale@v4.0.0
+    - uses: actions/stale@v5.1.0
       with:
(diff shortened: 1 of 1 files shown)


fmvilas (approved):
LGTM 👍

derberg:
It is a super new option, won't work without bumping the action version number -> https://github.com/actions/stale/releases/tag/v5.1.0

nevertheless, of course makes sense

smoya:
> It is a super new option, won't work without bumping the action version number -> https://github.com/actions/stale/releases/tag/v5.1.0

Good catch! I've added a new commit bumping the version.

derberg (approved):
/rtm

PR #175, opened 2022-07-29
feat: add error comment in lint-pr-title.yml

Krishks369:
I have added feature in the lint-pr-title.yml file to comment on the lint errors in the title while PR is created

 **Three scenarios are there for the workflow**

1.  When the PR title is not in the correct format i.e when it is apart from titles like `feat`, `fix` etc. It works like it has in this [PR](https://github.com/Krishks369/.github/pull/16) .

2. When the PR title is not found or the PR subject starts in capital, then it works like in this [PR](https://github.com/Krishks369/.github/pull/17)

3. When the lint is corrected after the comment, the comment is then deleted from the PR, as mentioned in https://github.com/marocchino/sticky-pull-request-comment

Although the action used for linting the title is https://github.com/amannn/action-semantic-pull-request  I have forked and used this feature before it got merged in the main action. So temporarily it uses https://github.com/Krishks369/action-semantic-pull-request.

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -25,7 +25,7 @@ jobs:
           subjectPatternError: |
             The subject "{subject}" found in the pull request title "{title}" should start with a lowercase character.
             
-      # Comments the error message from the above lint_pr_title action and deletes the comment when the linting is correct
+      # Comments the error message from the above lint_pr_title action 
       - if: always()
         name: Comment on PR
         uses: marocchino/sticky-pull-request-comment@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pr-title.yml
```diff
@@ -1,23 +1,46 @@
...
+          message: |
+                   ```
+                    ${{ steps.lint_pr_title.outputs.error_message}}
+                   ```
+       # deletes the error comment if the title is correct            
+      - if: ${{ steps.lint_pr_title.outputs.error_message == null }}
+        name: delete the comment
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:   
+          header: comment
```
derberg, line 43: you tested it, right?

derberg (changes requested):
I left some comments

derberg:
@Krishks369 hey man, need a hand?

sambhavgupta0705:
Hey @derberg ,do we need to refine this PR also?

PR #178, opened 2022-09-15
ci: fix link checker and stale bot

derberg:
- stale bot was staling issues and next day automatically unstaling - this is fixed with latest version
- link checker was in general behaving weird and throwing lots of false-positives

as you can see, I also started using commits as version pointers. We have global replicator, so we have luxury to use commits, instead invented GH Actions versioning, that is super insecure. I'm happy to explain more if needed


derberg:
@KhudaDad414 yo mate, please approve

KhudaDad414 (approved):

PR #180, opened 2022-09-20
ci: fix release workflow to support release branches

char0n:
Refs https://github.com/asyncapi/spec/pull/834
Refs https://github.com/asyncapi/spec-json-schemas/pull/255

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -9,12 +9,12 @@ on:
     branches:
       - master
       # below lines are not enough to have release supported for these branches
-      # make sure configuration of `semantic-release` package mentiones these branches
-      - next
+      # make sure configuration of `semantic-release` package mentions these branches
+      - next-spec
       - next-major
+      - next-major-spec
       - beta
       - alpha
-      - '**-release' # custom
 
 jobs:
 


derberg (approved):

char0n:
This was merged but haven't been propagated to https://github.com/asyncapi/spec/blob/master/.github/workflows/if-nodejs-release.yml and https://github.com/asyncapi/spec-json-schemas/blob/master/.github/workflows/if-nodejs-release.yml properly. Any advice?

derberg:
- for https://github.com/asyncapi/spec it did not as `if-nodejs-release.yml` is actually a legacy file that should not be there. Spec is its own beast, with dedicated [release workflow](https://github.com/asyncapi/spec/blob/master/.github/workflows/release.yml). Updates for `if-nodejs-release.yml` and other node related workflow no longer go there as `nodejs` topic is not there on the repo. And in general, there is nothing node-related with this repo
- for https://github.com/asyncapi/spec-json-schemas I actually see all is good, and change was propagated properly https://github.com/asyncapi/spec-json-schemas/pull/256/files

char0n:
Right, so I'll issue last PR that merges master to `next-spec` branch on `asyncapi/spec` repo and we should be fine there as well.

Issue #177, opened 2022-09-15
PR level check that verifies if contributed `package-lock.json` is created with `v2`


derberg:
We need a new global workflow, with a dedicated configuration in https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml that only certain projects can opt into if they want

The goal is that it runs on a PR and it checks if the PR modifies `package-lock.json` and the file is created with `"lockfileVersion": 2`. Otherwise the check fails and we post a nice friendly comment saying "oy, sorry but we use lockfileVersion 2, so please make sure you are not using npm 6 to install dependencies in the project, thanks a 💯 " (just an example 😄 ).

Then once the fix is done, we remove the bot comment

related to https://github.com/asyncapi/parser-js/issues/427

smoya:
@derberg FYI I created an issue in every repository that still sticks to package-lock version 2.

https://github.com/asyncapi/spec-json-schemas/issues/272
https://github.com/asyncapi/asyncapi-react/issues/652
https://github.com/asyncapi/parser-js/issues/652
https://github.com/asyncapi/extensions-catalog/issues/105
https://github.com/asyncapi/converter-js/issues/147
https://github.com/asyncapi/generator/issues/853
https://github.com/asyncapi/tck/issues/159
https://github.com/asyncapi/studio/issues/449
https://github.com/asyncapi/html-template/issues/380
https://github.com/asyncapi/markdown-template/issues/301
https://github.com/asyncapi/github-action-for-generator/issues/260
https://github.com/asyncapi/generator-hooks/issues/98
https://github.com/asyncapi/vs-asyncapi-preview/issues/129
https://github.com/asyncapi/shape-up-process/issues/111
https://github.com/asyncapi/chatbot/issues/69
https://github.com/asyncapi/create-glee-app/issues/135

derberg:
- we need to be flexible, so if someone wan't to use package-lock v2 and not v3 that is fine. So the check that we roll out must support validation of any version
- we need a composite reusable action that checks package-lock version from the default branch, and verifies it against version coming from the PR
- now problem is that what if someone intentionally want to switch from v2 to v3? CI will always complain. So we need to make sure that maintainer of given repo can say, "yo, my package lock version for my repo is v2". I think this can be only done through GitHub repo variables:
[...]

PR #185, opened 2022-10-10
ci: add check for package-lock.json version

smoya:
**Description**

This PR adds an extra check in case `package.json` is found that will check the `package-lock.json` version is set to `2`. Otherwise, the action will fail.

This avoids issues like https://github.com/asyncapi/parser-js/pull/650.

**Related issue(s)**
https://github.com/asyncapi/parser-js/pull/650

### DIFF
--- a/.github/workflows/if-nodejs-pr-testing.yml
+++ b/.github/workflows/if-nodejs-pr-testing.yml
@@ -34,6 +34,10 @@ jobs:
         run: test -e ./package.json && echo "::set-output name=exists::true" || echo "::set-output name=exists::false"
         shell: bash
       - if: steps.packagejson.outputs.exists == 'true'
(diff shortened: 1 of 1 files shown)


derberg:
@smoya hey, we cannot solve it this way. Please have a look -> https://github.com/asyncapi/.github/issues/177

It should be a separate workflow IMHO (not enforce it all suddenly on all projects that still use v1) + we should drop a comment to the PR instead of asking folks to go to logs and figure that issue might be related to a bad package-lock version.  Something we do now with PR title (and we do it as many people were asking for help instead of going to logs)

derberg:
@smoya you want to continue working on this one?

derberg:
I'll close it for now as we need different solution for it, so completely new PR
Will update https://github.com/asyncapi/.github/issues/177

Issue #123, opened 2022-01-14
Update Node and npm to a more recent version


smoya:
## TL;DR;
- Node and Npm version should be updated to a more recent one, as we are mostly developing with newer node and npm versions, also because committed lockfiles are version 2.
Lockfile version 2 is available starting at npm v7, However I think we could update it even to a more recent version.
For example, Node v16, that includes npm v8). 
- `npm install` command [here](https://github.com/asyncapi/.github/blob/24a01f622f206047c23108a6d234c01199efb61d/.github/workflows/if-nodejs-release.yml#L69) should be replaced by `npm ci` so we avoid side effects (no modification of package-lock.json).

Any strong reason to not do this?

## Long version if you are not very convinced

The following is a real use case of a bug in the produced image. It took me like an hour to figure out the "bug" (thanks to @magicmatatjahu for pairing!).

The very first Docker image of [server-api](https://hub.docker.com/layers/asyncapi/server-api/0.1.0/images/sha256-36321bde2f5068bcaed3226b46408b3ff0ce814c791c56aa850626c1cffa9255?context=explore) was built and pushed to Docker Hub by our GH workflow action [if-nodejs-release.yml](https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.yml).

Unfortunately, after trying to deploy it to a K8s cluster, we found out that the container was failing due to a missing node module: `Error: Cannot find module '@babel/core'`. By pulling the same image locally, we discarded the problem was in K8s but in the image itself.
The surprise is that, whenever that image is built locally (e.g. in my computer), the container works like a charm.

But why?

After diving (literally, using [dive](https://github.com/wagoodman/dive) tool) into both Docker images (the one created by CI and the one created by myself locally), I noticed there was a big difference on a particular Docker layer. In particular, in the one created by `COPY package* ./`.

![Diff images using dive](https://user-images.githubusercontent.com/1083296/149553068-0956db52-85ab-4a3c-90e4-fe33ac0e0592.png)

As you can see, the `package-lock.json` filesize differs. The one created by CI is much smaller (554 kB VS 1.2 MB).

After reading both files, I noticed that the `package-lock.json` from the image created by CI was a version 1 (`"lockfileVersion": 1`), however both the one created locally and the one available in the repository source code are version 2 (`"lockfileVersion": 2`).

[...]

PR #187, opened 2022-10-18
feat: install Node v16 when lockfileVersion is 2

smoya:
**Description**
Partially fixes https://github.com/asyncapi/.github/issues/123

Note: Not sure if this can be extracted to another action and reused?

**Related issue(s)**
- https://github.com/asyncapi/.github/issues/123
- https://github.com/asyncapi/parser-js/issues/427


derberg:
I think this is actually the smartest solution we could do here 🍺 . So setting up Node environment based on `lockfileVersion` and maybe other prerequisites too.

I'm only not sure if the way you do it is proper:
- there is a lot of repetition of the same command and yeah, we know what repetition ends up in future
- current script assumes there are just 2 lockfile versions. Thing is that since `npm` got acquired by GitHub, I have a feeling that development got faster, and they will evolve to more versions I think. In official docs you even have mention of lockfile `v3` -> https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json

sounds like best would be to have a custom GitHub action responsible for detecting what `package-lock` version is used and exposing it to output. It wouldn't have to be published to marketplace, could be placed inside this repo (example from generator -> https://github.com/asyncapi/generator/tree/master/.github/templates-list-validator and https://github.com/asyncapi/generator/blob/master/.github/workflows/templates-list-validation.yml#L14)

thoughts?

of course there is one downside, that we for example say that generator works with node > 12 but we will test on 16 only 😞 but these issues with `lockfileVersion` compatibility are out of our hands

we are not the only ones suffering https://twitter.com/search?q=lockfileVersion&src=typed_query&f=live

alequetzalli (approved):

smoya:
@derberg  should we move forward with this?

derberg:
@smoya did you see my comment?

other stuff:
- package lock v3 is here 😄 so we definitely need to add it to this PR too 😄 v3 and node 18
- please use `actions/setup-node@v3` instead of `actions/setup-node@v2`
- I don't think we can use `"::set-output name=lockfileVersion::1"` as according to https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ it will stop working soon. I recommend to use `actions/github-script@v4` action instead of bash. Thanks to it, you can use `core` library that is intend driven and their `setOutput` is future-proof. Also advantage of script is that you can put script [in one file](https://github.com/asyncapi/.github/tree/master/.github/workflows/scripts) and then reuse in 3 workflows, so you kinda address my concern from previous comment

smoya:
> @smoya did you see my comment?

Nope 😆 For some reason, I did not. Thanks for taking the time to collect and summarize all the info.
I like all your suggestions, except I should write javascript instead of a simple shell script. It's just a grep! We can still redirect an output to the right env vars: `echo "{name}={value}" >> $GITHUB_OUTPUT`.
I like the idea of having all of that into just one single script and reuse it. In fact, I think the script will just return the node version we should use, something like a `recommended-node-version-by-lockfile.sh` or similar.

derberg:
yeah, for you grep for me magic 😃 

I do not mind `.sh` as long as we just take this reusability direction. Global workflow can also push out `.sh` files to other repos

derberg:
when you will work on this, refactor I mean, you might want to try https://github.com/apache/airflow/pull/27371

more on composite actions -> https://asyncapi.slack.com/archives/C34AUKWQK/p1680613343254339

derberg:
closing in favour of https://github.com/asyncapi/.github/pull/230

Issue #191, opened 2022-10-25
Add automation to send email to TSC newsletter subscribers


derberg:
We need to refactor the automation that we have for sending emails to subscribers that want to learn about new meetings.

Automation should also support sending email to new kind of subscribers, that want to know about any time TSC needs to vote.

We need to integrate new Mailchimp flow to already existing automation that we have for TSC when we drop message to slack

core code info:
```
            recipients: {
                list_id: '6e3e437abe',
                segments_opts: {
                    saved_segment_id: 'tsc-voting-email'
                }
            }
```

PR #192, opened 2022-10-27
ci: add email sending to TSC notify workflow and new scripts concept

derberg:
**Description**

- we know this new functionality works as it was first merged only in community repo -> https://github.com/asyncapi/community/commit/acb42a92e89c64bca9c0d8a92057fd532376e4c8
- this is how email looks like. Link mentioned there in screen shot points [here](https://asyncapi.us12.list-manage.com/track/click?u=3cdae6aed726d2e7490aec739&id=9595ebe4f6&e=35b74043b1)
    Screenshot 2022-10-27 at 10 59 47
- you see new concept of globally managed scripts that will also be replicated. It will make it easier for us to make more rich workflows with reusable code. Another use case if for https://github.com/asyncapi/.github/pull/187 (cc @smoya ). Scripts are basically easier than creating custom actions.

**Related issue(s)**
Resolves https://github.com/asyncapi/.github/issues/191
See also https://github.com/asyncapi/community/issues/253

### DIFF
--- a/.github/workflows/global-replicator.yml
+++ b/.github/workflows/global-replicator.yml
@@ -86,7 +86,7 @@ jobs:
 
   replicate_generic_workflows:
     if: startsWith(github.repository, 'asyncapi/')
-    name: Replicate generic workflows needed for any project
+    name: Replicate generic workflows and scripts needed for any project
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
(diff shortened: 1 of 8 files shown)


KhudaDad414 (approved):
LGTM! nitpicking here but IMO the name `mailchimp` doesn't show what kind of script it is holding. maybe we can move the scripts inside `mailchimp/email-tsc` or rename `mailchimp` to `email-tsc`?

derberg:
@KhudaDad414 I intentionally did `mailchimp/index.js` with just a code comment so later, when we come up with more use cases to send emails (I'm pretty sure we will) so then we refactor, add more folders and split code and just keep `index.js` as main entry. `email-tsc` is super specific about TSC, but after refactoring we can send emails for other use cases, not only tsc related. You ok if we leave it as it is now?

Issue #194, opened 2022-11-14
Improve email sent to TSC members to include the title of the topic


derberg:
Now email sent to TSC is always the same, just the link to related topic is different. But link is hidden behind the same display title `Click here to see more details on GitHub`

We need to improve it. Have something like `Check on GitHub:` and then put a link with display name that is the same as the title of the issue/pr/discussion

Hints:
- this is where content of email is defined https://github.com/asyncapi/.github/blob/master/.github/workflows/scripts/mailchimp/htmlContent.js#L389
- `htmlContent.js` function needs to be extended to not only support `link` argument but also `title` (same here https://github.com/asyncapi/.github/blob/master/.github/workflows/scripts/mailchimp/index.js#L50 and here https://github.com/asyncapi/.github/blob/master/.github/workflows/scripts/mailchimp/index.js#L13)
- every time `sendEmail` function is called in [this workflow](https://github.com/asyncapi/.github/blob/master/.github/workflows/notify-tsc-members-mention.yml) we should pass the title, like for example instead of `sendEmail('${{github.event.issue.html_url}}');` we need `sendEmail('${{github.event.issue.html_url}}', '${{github.event.issue.title}}');`

I'm also not sure, but might be that we should include the title in email subject 🤔

PR #196, opened 2022-11-22
fix: improve email notifications for tsc voting

derberg:
**Description**

- title of issue/pr/discussion will now be included in body of email and the email title
- action will not react on edits as already because of it, few times we sent duplicated emails to people. We need to avoid to much spam

**Related issue(s)**
Resolves https://github.com/asyncapi/.github/issues/194

### DIFF
--- a/.github/workflows/scripts/mailchimp/index.js
+++ b/.github/workflows/scripts/mailchimp/index.js
@@ -10,7 +10,7 @@ const htmlContent = require('./htmlContent.js');
  * Sending API request to mailchimp to schedule email to subscribers
  * Input is the URL to issue/discussion or other resource
  */
-module.exports = async (link) => {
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/mailchimp/index.js
```diff
@@ -32,7 +32,7 @@ module.exports = async (link) => {
                 }
             },
             settings: {
-                subject_line: 'AsyncAPI TSC members attention required',
+                subject_line: `TSC attention required: ${ title }`,
```
derberg, line 35: we do not need to mention `AsyncAPI` in email subject as email is sent from AsyncAPI anyway. Shorter title, more chance to show more characters from the discussions title

KhudaDad414 (approved):
LGTM!

Issue #184, opened 2022-09-29
we should add some custom filter to release workflow


derberg:
#### Reason/Context

Release https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.yml relies on `semantic-release` to answer if release is needed.

This means, on every single commit it runs tests, and then release just to realize at the very end that it was not needed.

Bit waste of resources!

#### Description

I think we need to make it skip early all jobs if we notice that PR title is different than `feat`, `fix`, `feat!`

what we loose: sometimes (1% of cases) it happens that some releases of features fail, and that another `chore` commit fixes it, as release runs cause semantic release discovers unreleased feature. It will obviously no longer work as semantic-release will not ever run with this new solution

Thoughts?

smoya:
+1 to run **only** if the current commit has any of those prefixes (`feat`, `fix`, `feat!`).

> what we loose: sometimes (1% of cases) it happens that some releases of features fail, and that another chore commit fixes it, as release runs cause semantic release discovers unreleased feature. It will obviously no longer work as semantic-release will not ever run with this new solution

If that means we will need to add an extra commit with those prefixes in order to run the release, I'm fine to be honest.

PR #198, opened 2022-11-24
ci: improve release workflows to skip early

derberg:
**Description**

- this change will significantly improve GitHub Actions workload as at the moment release workflow is triggered on every single commit, and tests have to go through until `semantic-release` finds out release is not needed
- I use `startsWith(fromJson` and not `contains(fromJson` like in other workflows because commit message can be multiline, and we do not have 100% assurance that PR was not manually merged and feat/fix was not mentioned somewhere in the message
- I checked if `startsWith` also works with `contains` as there is not direct info in docs, at least I did not find it. I just searched github and found https://github.com/insignias/React/blob/main/.github/workflows/merge-release-pr-to-main.yml#L40

**Related issue(s)**
Resolves https://github.com/asyncapi/.github/issues/184

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -22,7 +22,12 @@ jobs:
     # We just check the message of first commit as there is always just one commit because we squash into one before merging
     # "commits" contains array of objects where one of the properties is commit "message"
     # Release workflow will be skipped if release conventional commits are not used
-    if: ${{ !startsWith(fromJson('["fix:", "feat:", "fix!:", "feat!:"]'), github.event.push.commits[0].message) }}
+    if: |
+      startsWith( github.repository, 'asyncapi/' ) && 
+      (startsWith( github.event.commits[0].message , 'fix:' ) ||
+      startsWith( github.event.commits[0].message, 'fix!:' ) ||
+      startsWith( github.event.commits[0].message, 'feat:' ) ||
+      startsWith( github.event.commits[0].message, 'feat!:' ))
     name: Test NodeJS release on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -19,6 +19,10 @@ on:
 jobs:
 
   test-nodejs:
+    # We just check the message of first commit as there is always just one commit because we squash into one before merging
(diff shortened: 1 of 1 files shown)


fmvilas (dismissed):

.github/workflows/if-nodejs-release.yml
```diff
@@ -19,6 +19,15 @@ on:
 jobs:
 
   test-nodejs:
+    # We just check the message of first commit as there is always just one commit because we squash into one before merging
+    # "commits" contains array of objects where one of the properties is commit "message"
+    # Release workflow will be skipped if release conventional commits are not used
+    if: |
+      startsWith( github.repository, 'asyncapi/' ) &&
```
derberg, line 26: to make sure workflow does not run on forks, to not confuse people

I already had few questions from folks asking why release is failing (but it was running on their forks)

.github/workflows/if-nodejs-release.yml
```diff
@@ -19,6 +19,15 @@ on:
 jobs:
 
   test-nodejs:
+    # We just check the message of first commit as there is always just one commit because we squash into one before merging
+    # "commits" contains array of objects where one of the properties is commit "message"
+    # Release workflow will be skipped if release conventional commits are not used
+    if: |
+      startsWith( github.repository, 'asyncapi/' ) && 
+      (startsWith( github.event.commits[0].message , 'fix:' ) ||
```
derberg, line 27: unfortunately `github.event.commits[0].message` must be repeated several times. When I added it as global env in  workflow, it was not accessed in condition

.github/workflows/if-nodejs-release.yml
```diff
@@ -19,6 +19,15 @@ on:
...
+      startsWith( github.event.commits[0].message, 'fix!:' ) ||
+      startsWith( github.event.commits[0].message, 'feat:' ) ||
+      startsWith( github.event.commits[0].message, 'feat!:' ))
```
derberg, line 30: unfortunately `startsWith` has to be done this way, it accepts only `string` as input for evaluation. In case of `contains` it accepts array.

derberg (commented):
@fmvilas sorry, please approve again. When I was in the sauna yesterday, I was thinking about this workflow and something told me to do some tests on the personal repo. And yeah, as a result, I had to rework the entire condition

fmvilas (approved):
Yeah, that makes sense. I mean, the PR, not the sauna thing 😄

Issue #172, opened 2022-07-18
Modify release pipeline to install dependencies in a workflow directly and not through package.json


derberg:
#### Reason/Context

Comes from https://github.com/asyncapi/generator/issues/801

tl;dr

Installation of `semantic-release` and other release dependencies on a CI level is already practiced in our Go projects and works like a charm. Just have a look at:
  - https://github.com/asyncapi/parser-go/blob/master/.github/workflows/.releaserc - release config do not have to be in package.json and can be a separate file, inside `.github/workflows` directory, so we can replicate it automatically as other workflow files in all repos. 
  - https://github.com/asyncapi/parser-go/blob/master/.github/workflows/release.yml#L51 - since few releases of `semantic-release`, dependencies that we now explicitly add as a dependence (like `@semantic-release/github` ) are actually included in `semantic-release`. So we only need to explicitly install only `conventional-changelog-conventionalcommits`
  - https://github.com/asyncapi/parser-go/blob/master/.github/workflows/release.yml#L61) - we do not really even have to install `semantic-release` as a dependency but just use npx like `npx semantic-release` (just locking to specific version)

> there is one thing I will have to test before I proceed is, what if there is a repo that will get new config but already has some dependencies in place and also a release config in `package.json`. For sure dependencies used in a CI will be used instead of those installed with `package.json` but what about the config of the release, which one is first, `.releaserc` or `package.json` - that I do not know yet

#### Description

What I do with this task is:
- introduce global release config for all. It will be replicated to all repos using `topics_to_include: common-release`.
  - **I will** add new topic to all repos that have `release` property in `package.json`, unless I know that like in case [parser-js](https://github.com/asyncapi/parser-js/blob/master/package.json#L131-L132) release steps are a bit custom
  - **I will not** modify `package.json` (remove `release` prop) for the repos that will start using global `.releaserc`. It is just to much effort. This should be done by maintainers at some point of time. Maybe I will find some time to run a script that in all these repos will create a `good first issue` issue to modify `package.json`
- modify current nodejs related release pipeline:
  - **I will** make sure we install only `conventional-changelog-conventionalcommits` with specific version, and run `npx semantic-release` (of specific version)  instead of `npm run release` that calls `semantic-release`
  - **I will not** modify `package.json` (remove `release` script). It is just too much effort. This should be done by maintainers at some point of time. Maybe I will find some time to run a script that in all these repos will create a `good first issue` issue to modify `package.json`

[...]

PR #204, opened 2022-12-15
ci: simplify release workflow configuration

derberg:
**Description**

- using ` --if-present` with all npm scripts. Before release configuration was a problem as you could easily forget to add some script that was just echoing message that script is not yet implemented. Now release configuration per repo will be simpler, you add scripts only if you really need them
- remove the need to have all these different dev dependencies in package json
   - this was generating lots of conflicts with libraries using the same packages as peer dependency. Also keeping track on versions of plugins and their match with `semantic-release`
   - now we are installing dependencies inside release workflow, and only specific versions that we know always work.
   - as you can see we basically now install just one plugin and only `semantic-release`. It is because semantic release is now shipped with many default plugins included
   - this change that will go to all nodejs repos has been tested here https://github.com/asyncapi/generator/pull/815
- remove boilerplate release configuration from `package.json`. Until now every maintainer had to add it manually. Now we keep config in `.releaserc` and can push it out as global file to any repo that needs it.

---
## Breaking Changes

tl;dr After we merge this PR, changes in `if-nodejs-release.yml`, `if-nodejs-version-bump.yml`, `bump.yml` files no longer will be replicated to all nodejs projects

> it doesn't mean they will stop working, it's just that future improvements will not be pushed there

#### Why?

`semantic-release` uses https://github.com/davidtheclark/cosmiconfig for loading configurations, and according to their docs, unfortunately, first priority is package.json and later dedicated files 😞 In theory it should not be a problem, but because of different dependencies on package.json level and the ones installed in the workflow...I just do not want to "assume all will be fine"

Also, the cleanup anyway has to be done in the repo before new workflows are applied.

#### What will maintainers have to do?

There will be instructions, but basically they:
- need to cleanup `package.json` from dependencies, and release config
- add proper topics to reposiotory
- once merged, they need to ask `.github` repo maintainers to run workflow that will push new workflows to the repo

---
what do you think?

**Related issue(s)**
Resolves https://github.com/asyncapi/.github/issues/172


.github/workflows/global-replicator.yml
```diff
@@ -78,12 +78,48 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Replicating file
+        uses: derberg/copy-files-to-other-repositories@v1
+        with:
+          github_token: ${{ secrets.GH_TOKEN }}
+          patterns_to_include: .github/workflows/if-nodejs-release.yml,.github/workflows/if-nodejs-version-bump.yml,.github/workflows/bump.yml
+          topics_to_include: get-global-node-release-workflows
```
derberg, line 99: I was thinking that we can introduce such pattern for topics for future, so they start with `get-global` prefix

what do you think? better ideas

example: 👇🏼 
Screenshot 2022-12-15 at 17 09 11

derberg:
@KhudaDad414 can you have a look? these topics names 🤔 


@jonaslagoni @magicmatatjahu @smoya pinging you so you are aware of this proposal. That with merge of this PR, auto updates of release and other workflows will stop until repo owner opts in again. Advantage in your cases is modelina next releases and parser/parser plugin releases. So you can easily modify `bump` workflows there as you need for your cases. Then once you are done, you can opt in again. Aaaaaaand long term this will be a feature, that on specific occasion you can just opt-out from release workflows updates whenever you want by just removing the topic. Does it make sense?

derberg:
@KhudaDad414 dude you expect too much from me. I asked ChatGPT for you but even AI provides stupid answer 🤣 

Screenshot 2023-01-13 at 11 05 37

On the serious side. I know https://github.com/nektos/act is there and is very popular, but I was too lazy to try it out 😄 

I have 2 approaches, also because in complex scenarios I do not trust tools to do proper integration tests for me, and I value manual testing more (this is why there are still 100s of companies still having big teams of manual testers)

- I have separate test GH account where I manually test big stuff, and main repo is https://github.com/lukasz-lab/.github. I use it in majority of cases for testing the global workflow features. And I also have a repo on my main profile where I ran tests https://github.com/derberg/test-experiment/
- And what I do additionally now. If change can do some damage and I'm not sure of it. I first add it to one repo. In other words changes that I introduce in this workflow, were tested in production in https://github.com/asyncapi/generator/pull/815 first

Sorry for disappointing you 😄

KhudaDad414:
@derberg thanks mate for the explanation. 
I actually tried `act` but since our workflows are not that simple it let me down. 😆

Issue #199, opened 2022-11-28
Link checker having an insane amount of false positives


jonaslagoni:
#### Describe the bug
The link checker workflow: https://github.com/asyncapi/.github/blob/master/.github/workflows/link-check-pr.yml have an insane amount of false positives (where it reports an error when there is none).

False positives it detects:
- Relative references within the same markdown file `[something](#Something)`.
- Because we use https://allcontributors.org/ it can happen users have links to websites that is unreachable. These should never be checked as failing a workflow because of that makes no sense since we have no control over the link.
- `429` failing workflow when checking contributors based on allcontributors.

Example run from Modelina:
```
2022-11-28T06:34:13.2822535Z [0;33m=========================> MARKDOWN LINK CHECK <=========================[0m
2022-11-28T06:34:13.2849469Z 
2022-11-28T06:34:13.2849973Z FILE: API.md
2022-11-28T06:34:13.2854894Z   [✖] #AbstractGenerator → Status: 404
2022-11-28T06:34:13.2958901Z   [✖] #AbstractRenderer → Status: 404
2022-11-28T06:34:13.2959302Z   [✖] #AsyncapiV2Schema → Status: 404
2022-11-28T06:34:13.2959624Z   [✖] #CommonModel → Status: 404
2022-11-28T06:34:13.2959983Z   [✖] #Draft4Schema → Status: 404
2022-11-28T06:34:13.2960304Z   [✖] #Draft6Schema → Status: 404
2022-11-28T06:34:13.2960604Z   [✖] #Draft7Schema → Status: 404
2022-11-28T06:34:13.2960934Z   [✖] #InputMetaModel → Status: 404
2022-11-28T06:34:13.2961279Z   [✖] #OpenapiV3Schema → Status: 404
2022-11-28T06:34:13.2961593Z   [✖] #OutputModel → Status: 404
2022-11-28T06:34:13.2962218Z   [✖] #RenderOutput → Status: 404
2022-11-28T06:34:13.2962591Z   [✖] #SwaggerV2Schema → Status: 404
2022-11-28T06:34:13.2962972Z   [✖] #AsyncAPIInputProcessor → Status: 404
2022-11-28T06:34:13.2963338Z   [✖] #InputProcessor → Status: 404
2022-11-28T06:34:13.2963716Z   [✖] #JsonSchemaInputProcessor → Status: 404
2022-11-28T06:34:13.2964097Z   [✖] #OpenAPIInputProcessor → Status: 404
[...]

PR #219, opened 2023-03-27
fix: use latest global workflow and remove link checkers

derberg:
This PR:
- uses latest replicator -> https://github.com/derberg/manage-files-in-multiple-repositories/releases/tag/v2
- in theory, because `.github/workflows/link-check-cron.yml,.github/workflows/link-check-pr.yml` are removed from repo and config, bot should remove these check from all the repositories in AsyncAPI org

Fixes https://github.com/asyncapi/.github/issues/199

---

if someone wants to have these workflows in repository, they should open a separate PR to add these files as separate configuration, that someone can opt-in only by using some specific repository topic name


KhudaDad414 (dismissed):
☕

derberg (commented):
@KhudaDad414 another approval please

I forgot that `.github/workflows/link-check-cron.yml,.github/workflows/link-check-pr.yml` must stay there as otherwise workflow will ignore removal if there is no pattern that would notice it 🤦🏼 

not a big deal, can be removed later from config file if needed

KhudaDad414 (approved):
Let's release the beast 😆

Issue #41, opened 2021-03-18
Think about an action that periodicaly checks emotions on issues


derberg:
#### Reason/Context
I do not like, as a user, when I get into some project, I see there was some important issue for the community, but it is either closed or opened forever, but people do not write anything but only share their emojis, thumbs up and others.

The expectation is that maintainers will see that, and frustration arises.

But come one, how could maintainers even see every single issue and all those emotions if you do not approach them directly 🤷🏼 

#### Description
Please try answering few of those questions
- maybe a github action in every repo that calls API on every issue and PR in a given repo and checks emojis
- have some settings like, report only issues and PRs with more than 5 emotions or something like that
- ping us on slack as we do now with new issue or pr, or releases

derberg:
@KhudaDad414 look ☝🏼 it works 🚀

darp99:
Hello sir , @derberg 

How can i start work on this issue ?

derberg:
@darp99 hey! Most important questions what is your experience with GitHub Actions. I need to know so I can suggest how to start. This issue with also require you to use GitHub REST or GraphQL API. You do not need to know much about APIs though as they provide client library.

darp99:
Sorry i don't want to disappoint you. i haven't used Github Action but i am exploring it. I know the REST API very well
Please guide me. what i have to do. i will learn if something is missing and try to give my best

derberg:
> Sorry i don't want to disappoint you

no worries, you definitely did not disappoint me. I just wanted to understand how deep in details should I go.

My advice, for now, forget about GitHub Actions -> these are the easiest in this task. More important is to figure how to do this calculation with, either GitHub REST API or GraphQL API

From GitHub Action point of view what you need to know for your research is:
- such action, that measures emotions will run in every single repo separately, so you need to figure how to check emotions on issues, PRs and discussions on repo level
- and yeah I don't think anything more is needed 🤔 

@KhudaDad414 any advice from your side? you did something similar when you worked on the dashboard concept, the HOT-ness of issue.

KhudaDad414:
not much, I just wanted to add: the [Graphql repository query](https://docs.github.com/en/graphql/reference/objects#repository) can be used to parse all of the [reaction groups](https://docs.github.com/en/graphql/reference/objects#reactiongroup) of every PRs, issues and discussions with a single Graphql query 😀

sparsh-989:
Hey,
Can I work on this issue?

14Richa:
[...]

PR #220, opened 2023-03-28
ci:  new github action which logs all the reactions in prs and issues

14Richa:
**Description**
 A GitHub action that calls API on every issue and PR and labels `(community-attention)` in the issues and PR which have more than 5 reactions.

**Pre-requisite** 
Ensure the label we are trying to add ("more than 5 reactions") already exists in the repository. 
If it doesn't exist, you'll need to create it first using the GitHub web interface.

**Related issue(s)**

Resolves: #41

### DIFF
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -4,6 +4,9 @@ on:
     types: [opened]
   pull_request:
     types: [opened]
+  schedule:
+    - cron: "0 0 * * 0" # Run at midnight on Sundays
+  workflow_dispatch: {} # Allow manual triggering
 jobs:
   check-reactions:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -14,49 +14,42 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const { Octokit } = require('@octokit/rest');
-            const octokit = new Octokit({ auth: '${{ secrets.GITHUB_TOKEN }}'});
             const { owner, repo } = context.repo;
             // Check issues with more than 5 reactions
-            const issuesResponse = await octokit.issues.listForRepo({
+            const issues = await github.paginate(github.issues.listForRepo.endpoint({
               owner,
               repo,
               state: 'open',
-            });
-            if (issuesResponse.data) {
-              console.log(`Found ${issuesResponse.data.length} issues`);
-              for (const issue of issuesResponse.data) {
-                console.log(`Checking issue #${issue.number}`);
-                if (issue.reactions && issue.reactions.total_count > 5) {
-                  console.log(`Issue #${issue.number} has ${issue.reactions.total_count} reactions`);
-                  await octokit.issues.addLabels({
-                    owner,
-                    repo,
-                    issue_number: issue.number,
-                    labels: ['more than 5 reactions'],
-                  });
-                }
+            }));
+
+            for (const issue of issues) {
+              console.log(`Checking issue #${issue.number}`);
+              if (issue.reactions && issue.reactions.total_count > 5) {
+                console.log(`Issue #${issue.number} has ${issue.reactions.total_count} reactions`);
+                await github.issues.addLabels({
+                  owner,
+                  repo,
+                  issue_number: issue.number,
+                  labels: ['more than 5 reactions'],
+                });
               }
             }
-
             // Check pull requests with more than 5 reactions
-            const pullsResponse = await octokit.pulls.list({
+            const pulls = await github.paginate(github.pulls.list.endpoint({
               owner,
               repo,
               state: 'open',
-            });
-            if (pullsResponse.data) {
-              console.log(`Found ${pullsResponse.data.length} pull requests`);
-              for (const pull of pullsResponse.data) {
-                console.log(`Checking pull request #${pull.number}`);
-                if (pull.reactions && pull.reactions.total_count > 5) {
-                  console.log(`Pull request #${pull.number} has ${pull.reactions.total_count} reactions`);
-                  await octokit.issues.addLabels({
-                    owner,
-                    repo,
-                    issue_number: pull.number,
-                    labels: ['more than 5 reactions'],
-                  });
-                }
+            }));
+            
+            for (const pull of pulls) {
+              console.log(`Checking pull request #${pull.number}`);
+              if (pull.reactions && pull.reactions.total_count > 5) {
+                console.log(`Pull request #${pull.number} has ${pull.reactions.total_count} reactions`);
+                await github.issues.addLabels({
+                  owner,
+                  repo,
+                  issue_number: pull.number,
+                  labels: ['more than 5 reactions'],
+                });
               }
             }
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -1,0 +1,19 @@
+name: Check Reactions on Pr and Issue
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+jobs:
+  check-reactions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Check Reactions
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl -H "Authorization: token $GITHUB_TOKEN" \
+            "https://api.github.com/repos/${{ github.repository }}/issues?state=all" \
+            | jq '.[] | {number: .number, title: .title, reactions: .reactions}'
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -20,6 +20,7 @@ jobs:
               owner,
               repo,
               state: 'open',
+              per_page: 100,
             }));
 
             for (const issue of issues) {
@@ -39,6 +40,7 @@ jobs:
               owner,
               repo,
               state: 'open',
+              per_page: 100,
             }));
             
             for (const pull of pulls) {
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -1,7 +1,7 @@
 name: Check Reactions on Pr and Issue
 on:
   schedule:
-    - cron: "0 0 * * 0" # Run at midnight on Sundays
+    - cron: "0 13 * * 0" # Run at 1 PM on Sundays
   workflow_dispatch: {} # Allow manual triggering
 jobs:
   check-reactions:
--- a/.github/workflows/check-reactions-on-pr-and-issues.yml
+++ b/.github/workflows/check-reactions-on-pr-and-issues.yml
@@ -14,6 +14,32 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          curl -H "Authorization: token $GITHUB_TOKEN" \
+          run: |
+          echo "Checking reactions..."
+            issues=$(curl -H "Authorization: token $GITHUB_TOKEN" \
             "https://api.github.com/repos/${{ github.repository }}/issues?state=all" \
-            | jq '.[] | {number: .number, title: .title, reactions: .reactions}'
+            | jq '.[] | select(.reactions.total_count > 5) | .number')
+          echo "Found issues with more than 5 reactions: $issues"
+          for issue in $issues; do
+            echo "Adding label to issue #$issue..."
+            curl -X POST -H "Authorization: token $GITHUB_TOKEN" \
+              -H "Accept: application/vnd.github.v3+json" \
+              "https://api.github.com/repos/${{ github.repository }}/issues/$issue/labels" \
+              -d '{"labels": ["more than 5 reactions"]}'
+          done
+          echo "Checking reactions..."
+          prs=$(curl -H "Authorization: token $GITHUB_TOKEN" \
+          "https://api.github.com/repos/${{ github.repository }}/pulls?state=all" \
+          | jq '.[] | select(.reactions.total_count > 5) | .number')
+          echo "Found pr with more than 5 reactions: $prs"
+          if [ -z "$prs" ]; then
+            echo "No pull requests with more than 5 reactions found."
+          else
+            for pr in $prs; do
+              echo "Adding label to pr #$pr..."
+              curl -X POST -H "Authorization: token $GITHUB_TOKEN" \
+              -H "Accept: application/vnd.github.v3+json" \
+              "https://api.github.com/repos/${{ github.repository }}/pulls/$pr/labels" \
+              -d '{"labels": ["more than 5 reactions"]}'
+            done
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,65 @@
...
+  pull_request:
+    types: [opened]
+jobs:
+  check-reactions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm install @octokit/rest
```
derberg, line 14: did you have to install it? afaik `actions/github-script@v4` makes sure you have access to library by default and you do not have to pass auth secrets as it reads secrets by default 🤔

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,65 @@
...
+          script: |
+              const { Octokit } = require('@octokit/rest');
+              const octokit = new Octokit({ auth: '${{ secrets.GITHUB_TOKEN }}'});
+              const { owner, repo } = context.repo;
+
+              // Check issues with more than 5 reactions
+              const issuesResponse = await octokit.issues.listForRepo({
+                owner,
+                repo,
+                state: 'all',
```
derberg, line 28: how does it work? it is part of octokit that makes you pull all issues without paging?

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,65 @@
...
+                console.log(`Found ${issuesResponse.data.length} issues`);
+                for (const issue of issuesResponse.data) {
+                  console.log(`Checking issue #${issue.number}`);
+                  if (issue.reactions && issue.reactions.total_count > 5) {
+                    console.log(`Issue #${issue.number} has ${issue.reactions.total_count} reactions`);
+                    await octokit.issues.addLabels({
+                      owner,
+                      repo,
+                      issue_number: issue.number,
+                      labels: ['more than 5 reactions'],
```
derberg, line 40: I guess that if we add this label, we can also filter by this label, to not perform `addLabels` on all issues but only those that do not have a label?

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,65 @@
+name: Check Reactions on Pr and Issue
+on:
```
derberg, line 2: instead of reacting on event that issue or PR was created we need two triggers imho:
- cron, that runs once a week, I think enough
- workflow dispatch that allows for manual trigger in case someone do not want to wait until next cron trigger

thoughts?

14Richa: Sounds good to me. @derberg 

It was the first prototype so I simply triggered it with PR and issue. But definitely, we can add a cron job to the workflow.

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: Yes, you are correct. Adding a label to the issues and pull requests that have more than 5 reactions would allow us to filter.

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: @derberg, I have added the cron job which runs at midnight on Sundays, I have also enabled the workflow dispatch which triggers manually, and removed the earlier triggers.

.github/workflows/check-reactions-on-pr-and-issues.yml
derberg: I think you missed this question

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,72 @@
...
+            const issuesResponse = await github.issues.listForRepo({
+              owner,
+              repo,
+              state: 'all',
+            });
+            if (issuesResponse.data) {
+              console.log(`Found ${issuesResponse.data.length} issues`);
+              for (const issue of issuesResponse.data) {
+                console.log(`Checking issue #${issue.number}`);
+                const labelsResponse = await github.issues.listLabelsOnIssue({
```
derberg, line 28: oh, we need to query for list of labels, this info is not part of `issuesResponse.data` or accessible in some kind of easier way without additional API call?

asking case I waned us to filter by `!labels.includes('more than 5 reactions')` to limit numer of calls, and not increase them 😄

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Check Reactions
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { Octokit } = require('@octokit/rest');
+            const octokit = new Octokit({ auth: '${{ secrets.GITHUB_TOKEN }}'});
```
derberg, line 18: this is not needed, that is the whole beauty of using `actions/github-script`

https://github.com/actions/github-script says that out of the box you have `github A pre-authenticated [octokit/rest.js](https://octokit.github.io/rest.js) client with pagination plugins`

and then have a look at how you should do pagination -> https://github.com/actions/github-script#welcome-a-first-time-contributor

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  schedule:
+    - cron: "0 0 * * 0" # Run at midnight on Sundays
+  workflow_dispatch: {} # Allow manual triggering
+jobs:
+  check-reactions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
derberg, line 11: use v3 please, we will soon have to update all our workflows to use it too

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Check Reactions
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
derberg, line 15: is it still needed?

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,55 @@
...
+
+            for (const issue of issues) {
+              console.log(`Checking issue #${issue.number}`);
+              if (issue.reactions && issue.reactions.total_count > 5) {
+                console.log(`Issue #${issue.number} has ${issue.reactions.total_count} reactions`);
+                await github.issues.addLabels({
+                  owner,
+                  repo,
+                  issue_number: issue.number,
+                  labels: ['more than 5 reactions'],
```
derberg, line 33: `more than 5 reactions` I'm wondering if this is good label name, like what if we at some point of time say " maybe we should only monitor if there is more than 10 emotions

maybe we can have a better, and also "no spaces" label name?

derberg (commented):
few final questions

also, is there a repo where you test it that code works?

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: @derberg , how about `reactions-5plus` ?

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: Yes

.github/workflows/check-reactions-on-pr-and-issues.yml
derberg: what about my comment about problem with using specific number?

so basically have a label that do not refer to the amount of emotions but the fact that it requires attention as many community members are involved

.github/workflows/check-reactions-on-pr-and-issues.yml
derberg: but this is a generic token, the default one GitHub Actions have access to, and `github-scripts` claims that `github` client is preauthorised, so my assumption is that they access the token on their own from secrets, no?

their docs say:
```
Note that github-token is optional in this action, and the input is there in case you need to use a non-default token.
```

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: Few more options IMHO: 
1. `community-attention` 
2. `community-review` 
3. `community-discussion`

.github/workflows/check-reactions-on-pr-and-issues.yml
14Richa: In the docs, If we want to print the available attributes of context we don't need a token but if we want to comment on PR or add a label we do need `GITHUB_TOKEN`. Here is the link: https://github.com/actions/github-script/tree/releases/v4

Also in one of our [yaml](https://github.com/asyncapi/.github/blob/master/.github/workflows/welcome-first-time-contrib.yml#L22), we are using a token. 

If we don't use the token we are getting: `Error: Unhandled error: HttpError: Resource not accessible by integration`

derberg: ok, thanks for explanation 🙂

derberg:
@KhudaDad414 can you also have a look at this workflow?

I was thinking if maybe we should first add it to only one repo, to see how it works 🤔

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,55 @@
+name: Check Reactions on Pr and Issue
+on:
+  schedule:
+    - cron: "0 0 * * 0" # Run at midnight on Sundays
```
KhudaDad414, line 4: Not a fan of midnight. A random time would be a lot better.
```suggestion
    - cron: "0 13 * * 0" # Run at 1 PM on Sundays
```

.github/workflows/check-reactions-on-pr-and-issues.yml
thulieblack: This sounds cool, so if issues have a certain amount of emojis, they'll be added to the dashboard. Are there particular criteria for such issues or particular emojis that'll be used?

KhudaDad414 (changes requested):
Do we have an estimation of how many API calls will it make for all repos?

.github/workflows/check-reactions-on-pr-and-issues.yml
```diff
@@ -0,0 +1,55 @@
...
+  workflow_dispatch: {} # Allow manual triggering
+jobs:
+  check-reactions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Check Reactions
+        uses: actions/github-script@v4
```
KhudaDad414, line 13:
```suggestion
        uses: actions/github-script@v6
```

KhudaDad414:
@derberg I tested this in my fork and everything seems to be in order so I don't see the point. 🤔

.github/workflows/check-reactions-on-pr-and-issues.yml
derberg: @thulieblack dashboard would be basically a link to https://github.com/issues?q=is%3Aopen+org%3Aasyncapi+sort%3Aupdated-desc+label%3A%22community-attention%22+ for now. So list of all issues from all AsyncAPI. I do not know how to build a link that would point to all PRs

and it will list issues that have `community-attention` label. And this label will be added to every issue and PR that has more than 5 emojis (any) added. 

makes sense?

maybe we should increase to more than 10?

derberg: later we can extend https://www.asyncapi.com/community/dashboard maybe 🤔

thulieblack: Let's extend the emojis to more than 10. 
So for building a PR "dashboard" don't we need to use GitHub API to collect all PRs from all repos?

derberg: @thulieblack PR and issues, yes, definitely need to call the API. We have that logic in place already https://www.asyncapi.com/community/dashboard in `Hot Topics`. Just when we start adding `community-attention` labels, we should probably extend `Hot Topics` of just have another view. This is anyway not in scope for this PR. That is a separate discussion

14Richa:
> Do we have an estimation of how many API calls will it make for all repos?

Hey Khuda, yes we do have. 

One API call per repo, then as many calls as `issues with reactions > 10` we have in a repo, and the same with PR.

14Richa:
> @14Richa it is actually more. Single request fetch no more than 100 issues (I don't know what is the default per_page but max is 100) so in one repo like website it can be even 15 calls.

I agree. I guess we'd have to loop over the paginated results. 

> in UI I can query for issues that have a certain amount of reactions: https://github.com/asyncapi/website/issues?q=is%3Aissue+reactions%3A%3E5
so if REST API does not enable you to do this, maybe GraphQL API will?

I think the GitHub GraphQL API does not provide a direct way to filter issues based on the number of reactions.

>one small concern I have is that reactions that we change for now in this PR are the reactions on issue, but there can be reactions to the individual comment.

Agree with this. I modified the GitHub action on my [local repo to test it](https://github.com/14Richa/testRepo/blob/main/.github/workflows/ci-test1.yml#L38-L62). It works fine but it increases the API calls and the filtering still happens client side.

derberg:
> I agree. I guess we'd have to loop over the paginated results.

pagination plugin that you use there already handles that, just make sure page size is max

> Agree with this. I modified the GitHub action on my [local repo to test it](https://github.com/14Richa/testRepo/blob/main/.github/workflows/ci-test1.yml#L38-L62). It works fine but it increases the API calls and the filtering still happens client side.

yeah, I bet the amount of calls multiply a lot

> I think the GitHub GraphQL API does not provide a direct way to filter issues based on the number of reactions.

I just tried `search` and worked well:

```
query { 
  search(query: "org:asyncapi reactions:>10", type: ISSUE, first:100) { 
    issueCount
    edges{
      node {
        ... on Issue {
          url
        }
      }
    }
  }
}
```

result
```
{
  "data": {
    "search": {
      "issueCount": 6,
      "edges": [
        {
          "node": {
            "url": "https://github.com/asyncapi/website/issues/903"
          }
        },
        {
          "node": {
            "url": "https://github.com/asyncapi/cli/issues/251"
          }
        },
        {
          "node": {
            "url": "https://github.com/asyncapi/brand/issues/1"
          }
        },
        {
          "node": {
            "url": "https://github.com/asyncapi/spec/issues/618"
          }
        },
        {
          "node": {}
        },
        {
          "node": {
            "url": "https://github.com/asyncapi/spec/issues/94"
          }
        }
      ]
    }
  }
}
```

but now I'm thinking.... @KhudaDad414 aren't you actually doing it in your community dashboard for hot issues?

with such search query we do not need the action to run in all repos, but just in one. Still no solution for comments....

14Richa:
> I just tried search and worked well

I see, I didn't know about that way. I had tried and couldn't find a neat way to add filter from the nested `totalCount`.
```
query {
  search(type: ISSUE, query: "") {
    issueCount
    edges {
      node {
        number
        title
        reactions(first: 100) {
          totalCount
        }
      }
    }
  }
}

    ```

KhudaDad414:
@derberg, yes we are using `search` for the dashboard and it is awesome (very low points for api calls).

@14Richa have you checked https://github.com/asyncapi/website/blob/master/scripts/dashboard/issue-queries.js ? there are some queries that might help. also since you are going to query the reactions on the comments as well, it can consume more points which need to be tested.

PR #230, opened 2023-04-18
ci: node pipelines now setup node version depending on `lockfileVersion`

derberg:
This is related to very annoying thing about current pipelines, that they are set to Node 14 and in many projects package-lock v2 and even v3 is used, for newer versions of Node. As a result, many times happens that installation of project fail because of inconsistent support between old npm and new package lock structure.

more in https://github.com/asyncapi/.github/pull/187

so what is this PR introducing:
- composite action that can be reused across all repos and we can use it to verify version of package lock, and output proper node version that should be used
- modify pipelines that require Node setup to setup version of the node that given package-lock.json requires

### DIFF
--- a/.github/actions/get-node-version-from-package-lock/action.yml
+++ b/.github/actions/get-node-version-from-package-lock/action.yml
@@ -1,0 +1,43 @@
+name: 'Specify Node version basing on package-lock file version'
+outputs:
+  version:
+    description: 'Node.js version number to use with actions/setup-node action'
+    value: ${{ steps.getnode.outputs.version }}
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16
+        cache: 'npm'
+        cache-dependency-path: '**/package-lock.json'
+    - name: Get Node version
+      uses: actions/github-script@v6
+      id: getnode
+      with:
+        script: |
+          const { resolve } = require('path');
+          const packageLockLocation = './package-lock.json';
+          core.info(`Location of the packae-lock verification script is: ${ resolve(packageLockLocation) }`)
+          const packageLock = require(packageLockLocation);
+
+          const packageLockVersion = packageLock.lockfileVersion;
+          let nodeVersion;
+
+          switch (packageLockVersion) {
+            case 1:
+                nodeVersion = '14'
+                break;
+            case 2:
+                nodeVersion = '16'
+                break;
+            case 3:
+                nodeVersion = '18'
+                break;
+            default:
+                nodeVersion = '16'
+                break;
+          }
+
+          core.setOutput('version', nodeVersion);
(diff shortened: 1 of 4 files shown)


derberg:
@KhudaDad414 wdyt?

@smoya fyi

@jonaslagoni fyi, this is finally no going to be annoying anymore

KhudaDad414 (approved):
LGTM!

.github/actions/get-node-version-from-package-lock/action.yml
```diff
@@ -0,0 +1,37 @@
...
+            case 1:
+                nodeVersion = '14'
+                break;
+            case 2:
+                nodeVersion = '16'
+                break;
+            case 3:
+                nodeVersion = '18'
+                break;
+            default:
```
smoya, line 32: In which cases the default will apply? 
Do you know if there is a case when there is no `lockfileVersion` in the `package-lock.json`?
If there is not, it will mean the version is >3, so I would say then the default should be the greatest version of node?

.github/workflows/if-nodejs-pr-testing.yml
```diff
@@ -50,11 +50,15 @@ jobs:
         id: packagejson
         run: test -e ./package.json && echo "exists=true" >> $GITHUB_OUTPUT || echo "exists=false" >> $GITHUB_OUTPUT
         shell: bash
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Check package-lock version
+        uses: asyncapi/.github/.github/actions/get-node-version-from-package-lock@master
```
smoya, line 55: Question: Is this being cached? It will fetch latest action when updated in `master` branch?

smoya (approved):
Thanks @derberg for moving this forward! You rock!

Couple of nits and questions. Otherwise, LGTM! 🚀 🌔

.github/workflows/if-nodejs-pr-testing.yml
derberg: yup, taking latest from master, which is good, there are no issues with it, as it is in `.github` repo, so if we anyway change something to this composite action and it will be breaking, we will anyway update dependent workflows.

if there is a workflow that is not managed it `.github` repo and would like to use this composite action, this is fine, but then they need to use commitId and not branch name

.github/actions/get-node-version-from-package-lock/action.yml
derberg: it is hard to predict lockfile version evolution and what is latest node version. So in case `4` we set 16 and if it works, all fine, if it fails, it is actually good, we learn that there is a new version and we can fix it super fast here in the composite action

this default is only for projects that do not have package-lock, which is possible, by mistake, but possible

smoya (approved):

jonaslagoni (approved):
Awesome 👍

KhudaDad414 (approved):

PR #233, opened 2023-04-27
ci: update global workflow and cleanup release config file

derberg:
- new https://github.com/derberg/manage-files-in-multiple-repositories/releases/tag/v2.1.0
- new workflow that we can use specifically for removing files only
- update to have commit message always the same
- use static branch name, so one PR per repo and not `n` PR per repo depending how many jobs were triggered

all was driven by need to move a release config file from `.github/workflows` to root of repo that needs it


KhudaDad414 (approved):
Looks good from the outside.

PR #234, opened 2023-04-27
ci: skip workflows with new title and support skip for asyncapi-bot-eve

derberg:
- sometimes `asyncapi-bot-eve` autoupdates a PR and is last actor, so we need to also ignore it in some workflows
- use new pr title for global workflows


KhudaDad414 (approved):
Looks great.

PR #243, opened 2023-05-27
feat: added /review /rerequest command to get reviews from the reviewers

### LINKED ISSUE #211: We need another command for PRs which acts as re-request for review and also reminder for codeowners
We have cases like https://github.com/asyncapi/spec/pull/847#pullrequestreview-1331087602 when PR owner has hard time:
- pinging codeowners that did not yet perform any review on PR
- re-requesting review from people that already did a review, but the native re-request button feature in side bar do not work well

So we have already commands like, `ready-to-merge` or `autoupdate`. We should have more:
- `ping-for-attention`
- `rerequest`

quick proposals ☝🏼 that of course can be changed

---
before we talk about how to parse `codeowners` file, and then get data needed for requesting review, I suggest someone first check GitHub GraphQL or REST APIs as I think one day when I was exploring it, there was some API, probably GitHub Graphql mutation to trigger re-request review.

also maybe there are ready GitHub actions that we could use in workflows and there is no need to call API manually 🤔

Priyansh61:
**Description**
This action is used to call the reviewer to review the PR using the command /review
or /rerequest. 
- If the user is not mentioned, then all the reviewers are called to review the PR.
- If the user is mentioned, then only those users are called to review the PR.
- e.g.` /review @user1 @user2` this will call only user1 and user2 to review the PR
- e.g. `/review` this will call all the requested reviewers to review the PR

**Proof of Work**

https://github.com/asyncapi/website/assets/96513964/2b03d427-509f-4ef4-98f3-238c1bd84db7

**Related issue(s)**

Resolves #211


.github/workflows/request-review.yml
KhudaDad414, line 1: Can you please document these commands here as well? https://github.com/asyncapi/.github/blob/31b1ebe08b5ac7c0781bcaecc4ad768fc7d90436/.github/workflows/help-command.yml#L36

.github/workflows/request-review.yml
```diff
@@ -0,0 +1,76 @@
...
+                    const userIdNodeArray = [];
+                    for (const user of userIdArray) {
+                        const { data } = await github.request(`https://api.github.com/users/${user}`);
+                        const userId = data.node_id;
+                        userIdNodeArray.push(userId);
+                    }
+                    if (userIdNodeArray.length > 0) {
+                        const mutationQuery = `
+                        mutation { 
+                            requestReviews(input: {
```
KhudaDad414, line 63: some logs so it becomes easier to debug. :)

derberg:
@Priyansh61 do you plan to continue with this one?

derberg:
@Priyansh61 do you plan to continue with this one?

PR #246, opened 2023-06-26
feat: add pitch template

Amzani:
**Description**

As part of implementing ShapeUp methodology proposed in [this issue](https://github.com/asyncapi/studio/issues/659), I would like to create a new template for Pitch submissions.

This will simplify the creation of this kind of issues.

**Related issue(s)**

See also https://github.com/asyncapi/studio/issues/659

### DIFF
--- a/.github/ISSUE_TEMPLATE/pitch.yml
+++ b/.github/ISSUE_TEMPLATE/pitch.yml
@@ -1,0 +1,55 @@
+name: Pitch
+description: 'Do you have an idea and a product enhancement? Time to pitch!'
+title: '[PITCH]:'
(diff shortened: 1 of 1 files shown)


derberg:
@Amzani are you sure this should be in `.github` repo? it would be rolled out to all the other repos, while we already have https://github.com/asyncapi/.github/blob/master/.github/ISSUE_TEMPLATE/enhancement.md

ShapeUp is not widely adopted, just few folks use it

Amzani:
@derberg @KhudaDad414 you're right, I'll create this template only in studio repo then for now.

PR #248, opened 2023-07-12
ci: update MAINTAINERS.yaml based on CODEOWNERS changes

14Richa:
**Description**

This PR updates the MAINTAINERS.yaml file based on the changes made in the CODEOWNERS file. The workflow takes the repository name as input and adds it to the MAINTAINERS.yaml file.

When a new code owner is added, a corresponding maintainer object is created in the MAINTAINERS.yaml file with the repository name and the Twitter username of the code owner added to the object. It fetches the Twitter username using the GitHub API.

Similarly, when a code owner is removed, the corresponding maintainer object is removed from the MAINTAINERS.yaml file. The removal process checks if the maintainer has any other repositories before removing them from the MAINTAINERS.yaml file.

I have tested this workflow on my testRepo by adding a new code owner, and the MAINTAINERS.yaml file was updated with the new code owner object and repository name. Here are the logs from the workflow: https://github.com/14Richa/testRepo/actions/runs/5563061192/jobs/10161768757

Similarly, when a code owner is removed MAINTAINERS.yaml file was updated accordingly with the maintainer object removed from the file. Here are the logs from the workflow: https://github.com/14Richa/testRepo/actions/runs/5529277200/jobs/10087118152


.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,134 @@
...
+          fetch-depth: 2
+          ref: master
+          path: previous_state
+
+      - run: cd previous_state && git checkout HEAD^
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
derberg, line 33:
```suggestion
        uses: actions/setup-node@v3
        with:
          node-version: '16'
```

always latest please

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Compare CODEOWNERS
+        id: compare-codeowners
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('fs');
+            const yaml = require('js-yaml');
+
+            // Get repository name
+            const repoName = process.env.REPO_NAME;
```
derberg, line 53: sorry but why this, and also `Get repository name` step if you can do what you already did here: https://github.com/asyncapi/community/pull/750/files#diff-8ba49e6b37c137810bad8252fc74cb498dbeae5bc4650e947719905455ac7492R46

`const repo = context.repo.repo;`

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,134 @@
...
+            const updatedMaintainers = yaml.safeDump(maintainers);
+            fs.writeFileSync(maintainersFile, updatedMaintainers);
+            console.log('Updated MAINTAINERS.yaml:', updatedMaintainers);
+
+      - name: Commit and push
+        working-directory: ./current_state
+        run: |
+          git add .
+          git commit -m "Update MAINTAINERS.yaml"
+          git push https://${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} master
```
derberg, line 134: push alone will not work, you need to do the same as you did with this workflow: https://github.com/asyncapi/community/blob/master/.github/workflows/update-website-tsc.yml#L48-L51

create PR, as changes in repos can be done only through PRs

btw also makes sense to add slack alert

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,134 @@
...
+      - 'CODEOWNERS'
+
+jobs:
+  update-maintainers:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v2
```
derberg, line 16: version, consistency 😛 every time you copy/paste and do not check, a kitty dies 😄

derberg (changes requested):
yo, initial review for now, main algorithm review once we get basics fixed

.github/workflows/update-maintainers.yml
14Richa: Thanks for the reminder! :p 
Understood, I'll make sure to maintain consistency and double-check all details in the future.

KhudaDad414 (commented):
getting syntax error 
https://github.com/KhudaDad414/.github/actions/runs/5574259435/jobs/10182640765

14Richa:
> getting syntax error https://github.com/KhudaDad414/.github/actions/runs/5574259435/jobs/10182640765

Fixed it. Please check now!

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,161 @@
...
+              const regex = /@([a-zA-Z0-9_-]+)/g;
+              const matches = content.match(regex);
+              if (!matches) {
+                return [];
+              }
+              return matches.map(match => match.substr(1));
+            }
+      
+            const mainCodeowners = fs.readFileSync('./current_state/CODEOWNERS', 'utf8');
+            const prCodeowners = fs.readFileSync('./previous_state/CODEOWNERS', 'utf8');
```
KhudaDad414, line 64: Insn't these variables names a little confusing? PR and MAIN are basically the same here. can't we name these just like the branches? something like `previousCodeowners` and `currentCodeowners`

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,161 @@
...
+      
+            const mainUsernames = extractGitHubUsernames(mainCodeowners);
+            const prUsernames = extractGitHubUsernames(prCodeowners);
```
KhudaDad414, line 67: same here about naming...

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,161 @@
...
+              const twitterUsername = data.twitter_username;
+
+              // Exclude bot accounts
+              if (username === 'asyncapi-bot' || username === 'asyncapi-bot-eve') {
+                console.log('Skipping bot account:', username);
+                return; // Skip the iteration for bot accounts
+              }
+
+              const existingMaintainer = maintainers.find(maintainer => maintainer.github === username);
+              if (!existingMaintainer) {
```
KhudaDad414, line 101:
```suggestion
              const maintainerExists = maintainers.find(maintainer => maintainer.github === username);
              if (!maintainerExists) {
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,161 @@
...
+                maintainers.push({
+                  github: username,
+                  twitter: twitterUsername,
+                  isTscMember: false,
+                  repos: [repoName]
+                });
+                console.log('Added maintainer:', username);
+              } else {
+                console.log('Maintainer', username, 'already exists. Skipping addition.');
```
KhudaDad414, line 110: what about cases when maintainer does exists but they are being added as maintainer to another repo. shouldn't we add the current repo the their repo list?

KhudaDad414 (changes requested):
Isn't this PR supposed to open a pull request to the community repo? I see it applies the changes locally.

14Richa:
> Isn't this PR supposed to open a pull request to the community repo? I see it applies the changes locally.

Thank you for bringing this to my attention. I have updated the workflow. Now it creates a new branch and pushes the changes there. Subsequently, it opens a pull request for these changes to be reviewed and merged into the master branch in the community repo.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Install js-yaml
+        run: npm install js-yaml@3.14.1
+
+      - name: Install @octokit/rest
+        run: npm install @octokit/rest
+
+      - name: Compare CODEOWNERS
+        id: compare-codeowners
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
derberg, line 51:
```suggestion
          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+
+      - name: Install js-yaml
+        run: npm install js-yaml@3.14.1
+
+      - name: Install @octokit/rest
+        run: npm install @octokit/rest
+
+      - name: Compare CODEOWNERS
+        id: compare-codeowners
+        env:
```
derberg, line 50: I think you can do `env` on job level and then just reuse the `env` like in like 91 instead what we have in line 166

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+
+      - name: Install js-yaml
+        run: npm install js-yaml@3.14.1
+
+      - name: Install @octokit/rest
+        run: npm install @octokit/rest
```
derberg, line 46: why do you need to install it, it is for sure part of `actions/github-script@v6` action

Screenshot 2023-07-18 at 18 38 29

and rest client there is already preauthenticated so you do not need code like

```
const { Octokit } = require("@octokit/rest");

            const octokit = new Octokit({
              auth: process.env.GITHUB_TOKEN,
              request: {
                fetch: (url, options) => fetch(url, options)
              }
            });
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('fs');
+            const yaml = require('js-yaml');
+
+            // Get repository name
+            const repoName = context.repo.repo;
+
+            function extractGitHubUsernames(content) {
```
derberg, line 61: we need to make this regex "comments"-proof

some codeowners files have comments likekhttps://github.com/asyncapi/spec/blob/master/CODEOWNERS and there can be a situation that in the comment there is also `@` and a github handle, but commented out

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+              auth: process.env.GITHUB_TOKEN,
+              request: {
+                fetch: (url, options) => fetch(url, options)
+              }
+            });
+
+            // Update for added usernames
+            for (const username of addedUsernames) {
+              const { data } = await octokit.users.getByUsername({ username });
+              const twitterUsername = data.twitter_username;
```
derberg, line 100: better first, early just exclude bots, no need to call for twitter info from GH API if we have bot-usecase

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,181 @@
...
+                  console.log('Added repository to existing maintainer:', username);
+                } else {
+                  console.log('Maintainer', username, 'already exists and already has the repo. Skipping addition.');
+                }
+              }
+            }
+
+            // Update for removed usernames
+            for (const username of removedUsernames) {
+              const index = maintainers.findIndex(maintainer => maintainer.github === username);
```
derberg, line 130: why you need different approach here, other than in case of addition, why `findIndex`

derberg (changes requested):
some cleanup needed

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,172 @@
...
+          ref: master
+          path: previous_state
+
+      - run: cd previous_state && git checkout HEAD^
+
+      - name: Checkout community repo
+        uses: actions/checkout@v3
+        with:
+          repository: asyncapi/community
+          token: ${{ secrets.GITHUB_TOKEN }}
```
KhudaDad414, line 34: If you have to use a token, then please use the `GH_TOKEN` for consistency across this file.

KhudaDad414 (commented):
Looks great. 👍

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,172 @@
...
+                  repos: [repoName]
+                });
+                console.log('Added maintainer:', username);
+              } else {
+                // If the maintainer already exists, check if the current repo is in their list
+                if (!maintainer.repos.includes(repoName)) {
+                  maintainer.repos.push(repoName);
+                  console.log('Added repository to existing maintainer:', username);
+                } else {
+                  console.log('Maintainer', username, 'already exists and already has the repo. Skipping addition.');
```
derberg, line 114:
```suggestion
                  console.log(`Maintainer ${username} already exists and already has the repo. Skipping addition.`);
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,172 @@
...
+              const maintainer = maintainers.find(maintainer => maintainer.github === username);
+              if (maintainer) {
+                const repoIndex = maintainer.repos.indexOf(repoName);
+
+                if (repoIndex !== -1) {
+                  maintainer.repos.splice(repoIndex, 1);
+                  console.log(`Removed repository ${repoName} from maintainer ${username}`);
+
+                  if (maintainer.repos.length === 0) {
+                    const index = maintainers.indexOf(maintainer);
```
derberg, line 130: did you test it well and you sure it works fine

I usually use indexOf for simple types, and for objects it is not that obvious -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#comparing_objects

@KhudaDad414 thoughts?

derberg (commented):
2 comments only

.github/workflows/update-maintainers.yml
14Richa: @derberg, I did test it. Here are the logs: https://github.com/14Richa/testRepo/actions/runs/5610049235/jobs/10264525094
Codeowner was successfully removed from MAINTAINERS.yaml file.

derberg: but it is risky, as `indexOf` do not actually check the content but references, super risky in any future refactoring. I recommend changing to `findIndex`

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,171 @@
+name: Update MAINTAINERS.yaml
```
derberg, line 1:
```suggestion
name: Update MAINTAINERS.yaml in community repo
```

derberg (approved):
LGTM

before we merge, we need to manually add it to `generator` repo for testing

14Richa:
> before we merge, we need to manually add it to generator repo for testing 

@derberg , should I raise PR in the generator repo?

KhudaDad414 (approved):
LGTM!

derberg (dismissed):
lgtm

Issue #257, opened 2023-09-07
Evaluate github.com/rhysd/actionlint for linting GH Actions workflow files


smoya:
#### Reason/Context
At this moment, we run [the following validation action ](https://github.com/asyncapi/.github/blob/master/.github/workflows/validate-workflow-schema.yml) that runs on each PR and validates GH Action workflow files by validating the schema against the [following JSON Schema file](https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json).
That validation is very basic but still useful for catching very basic typos and errors.

However, yesterday I discovered https://github.com/rhysd/actionlint, a static linter for GitHub Actions workflow files. 

[The list of checks](https://github.com/rhysd/actionlint/blob/main/docs/checks.md) is massive, and I feel most of those are useful for our actions. For example, [checking variables](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-contexts-and-builtin-func), [syntax](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-syntax-expression) and [type checking](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-type-check-expression) for `${{}}` expressions, [dispatch events usage validation](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-workflow-dispatch-events), [shellcheck](https://github.com/koalaman/shellcheck) integration for commands in `run`, [dependencies validation](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-job-deps), [input checks](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-reusable-workflows) for reusable workflows, besides others.

I believe we could switch completely the action we use now in favor of this one and improve validation of GH Actions workflow files. This will help developers to develop new workflow files or modify existing ones without the concern of pushing changes with wrong content.

#### Description
Test https://github.com/rhysd/actionlint against some repo first, and see if it catches issues. I expect the linter will catch some errors on existing workflow files.

If we find those, it will mean the action is completely needed so I would recommend to include it in our pipelines.
In the other hand, if doesn't find errors, to decide if we want to switch or rather keep the simple validation we have now.

----

For code owners, please consider this as a `good-first-issue` @derberg @alequetzalli @fmvilas @KhudaDad414

mhmohona:
Hey @smoya, I would like to work on this.

smoya:
> Hey @smoya, I would like to work on this.

[...]

PR #268, opened 2024-01-14
fix: evaluated GitHub actions workflow files

mhmohona:
**Description**

- Updated GitHub Actions workflow to pass potentially untrusted data through environment variables as per GitHub's security hardening guidelines.
- Ensured that titles used in email notifications are now set via environment variables to prevent script injection attacks.

**Related issue(s)**

#257


.github/workflows/notify-tsc-members-mention.yml
```diff
@@ -66,10 +66,10 @@ jobs:
           CALENDAR_ID: ${{ secrets.CALENDAR_ID }}
           CALENDAR_SERVICE_ACCOUNT: ${{ secrets.CALENDAR_SERVICE_ACCOUNT }}
           MAILCHIMP_API_KEY: ${{ secrets.MAILCHIMP_API_KEY }}
+          ISSUE_TITLE: ${{ github.event.issue.title }}
         with:
           script: |
-            const sendEmail = require('./.github/workflows/scripts/mailchimp/index.js');
```
KhudaDad414, line 71: @mhmohona why are we removing the `sendEmail` import? 🤔

mhmohona:
I have a concern on it.

After running the actionlit for this repo it was the output - 
image

For first 2 errors, when I tested the scripts in https://rhysd.github.io/actionlint/, it didnt catch them.

If i address them, then both from actionlit online site and IDE I get error.

Should I address 1st 2 errors then? Or probably its a bug. :3

.github/workflows/validate-workflow-schema.yml
```diff
@@ -14,6 +14,16 @@ jobs:
...
+        run: |
+          curl -sLO https://github.com/rhysd/actionlint/releases/download/v1.6.3/actionlint_1.6.3_linux_amd64.tar.gz
+          tar xzf actionlint_1.6.3_linux_amd64.tar.gz
+          chmod +x actionlint
+          sudo mv actionlint /usr/local/bin/
+
+      - name: Run actionlint
+        run: actionlint -color
+
       - name: Install Dependencies
```
KhudaDad414, line 27: you can remove the rest of the steps. since `actionlint` will validate the file as well.

derberg (changes requested):
Amazing initiative, thanks!!!

Can you refactor to not download linter binary but use a dedicated github action 

https://github.com/raven-actions/actionlint

?

derberg:
@mhmohona do you plan to continue with this one?

derberg:
@mhmohona do you plan to continue with this one or should we ask community to continue?

.github/workflows/notify-tsc-members-mention.yml
```diff
@@ -291,7 +297,8 @@ jobs:
           CALENDAR_ID: ${{ secrets.CALENDAR_ID }}
           CALENDAR_SERVICE_ACCOUNT: ${{ secrets.CALENDAR_SERVICE_ACCOUNT }}
           MAILCHIMP_API_KEY: ${{ secrets.MAILCHIMP_API_KEY }}
+          DISCUSSION_TITLE: ${{ github.event.discussion.title }}
         with:
           script: |
             const sendEmail = require('./.github/workflows/scripts/mailchimp/index.js');
-            sendEmail('${{github.event.comment.html_url}}', '${{github.event.discussion.title}}');
+            sendEmail('${{github.event.comment.html_url}}', process.env.DISCUSSION_TITLE);
```
derberg, line 304: do you modify it because the linter complained?

I wonder then why `github.event.comment.html_url` did not have to be refactored as well 🤔

.github/workflows/validate-workflow-schema.yml
```diff
@@ -14,59 +14,5 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
```
derberg, line 15:
```suggestion
        uses: actions/checkout@v4
```

derberg (commented):
will you address the linter issues that are now shown in PR?

.github/workflows/validate-workflow-schema.yml
```diff
@@ -14,59 +14,5 @@ jobs:
...
-                } catch (error) {
-                  console.error(`Error validating file '${file}': ${error.message}`);
-                  process.exitCode = 1;
-                }
-              }
-            }
-            validateWorkflows();
\ No newline at end of file
+      - name: Check workflow files
+        uses: raven-actions/actionlint@v1
```
derberg, line 18: please do not use directly `v1` when action is not official, from github, but commit id (in your other PR you do it, so you have example there)

PR #277, opened 2024-03-14
ci: add workflow for the Bounty Program commands

aeworxet:
This PR adds GitHub Actions workflow to allow [Bounty Team members](https://github.com/orgs/asyncapi/teams/bounty_team) to issue commands to the organization's global AsyncAPI bot related to the Bounty Program, while at the same time preventing unauthorized users from misusing them.

This particular workflow adds a command that allows adding the label `bounty` to GitHub issues, thereby making them Bounty Issues.
Command is invoked on creation of comment `/bounty`.

### DIFF
--- a/.github/workflows/bounty-program-commands.yaml
+++ b/.github/workflows/bounty-program-commands.yaml
@@ -1,0 +1,89 @@
+# This workflow is centrally managed at https://github.com/asyncapi/.github/
+# Don't make changes to this file in this repository, as they will be overwritten with
+# changes made to the same file in the abovementioned repository.
+
+# The purpose of this workflow is to allow Bounty Team members
+# (https://github.com/orgs/asyncapi/teams/bounty_team) to issue commands to the
+# organization's global AsyncAPI bot related to the Bounty Program, while at the
+# same time preventing unauthorized users from misusing them.
+
+name: Bounty Program commands
+
+on:
+  issue_comment:
+    types:
+      - created
+
+jobs:
+  guard-against-unauthorized-use:
+    if: >
+      github.actor != ('aeworxet' || 'thulieblack') &&
+      (
+        contains(github.event.comment.body, '/bounty' )
+      )
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/bounty-program-commands.yaml
```diff
@@ -0,0 +1,89 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: ❌ @${{github.actor}} made an unauthorized attempt to use a Bounty Program's command
+        uses: actions/github-script@v6
+
+        with:
+          github-token: ${{ secrets.GH_TOKEN }}
+          script: |
+            const commentText = `❌ @${{github.actor}} is not authorized to use a Bounty Program's command.`;
```
derberg, line 34: maybe we should mention you and Thulie in comment
- so it is clear who is authorized
- so you are automatically pinged and see that people want to use it

.github/workflows/bounty-program-commands.yaml
```diff
@@ -0,0 +1,89 @@
...
+          {"name": "bounty", "color": "0e8a16", "description": "Participation in the Bounty Program"}
+        ]
+
+    steps:
+      - name: Add label `bounty`
+        uses: actions/github-script@v6
+
+        with:
+          github-token: ${{ secrets.GH_TOKEN }}
+          script: |
```
derberg, line 64: shouldn't we get a confirmation comment from bot here as well, that bounty addition was successful?
or is it enough that basically label will be added?

aeworxet: It will be observed in real-time whether the label is added or not; nothing more is needed or expected from the bot.

derberg (commented):
lgtm, please add it to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L130 as this is a workflow that must end up in every repository.

and yeah, please when you add a change, do not do force push, it is not helping in review 😅

aeworxet:
> please add it to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L130

- renamed `.yaml` to `.yml`
- added

derberg (approved):

Issue #278, opened 2024-03-15
if-nodejs-pr-testing.yml does not test on Windows 


jonaslagoni:
Problem workflow: https://github.com/asyncapi/.github/blob/3b278694c3d0850004e1e0e458779252195a5058/.github/workflows/if-nodejs-pr-testing.yml#L19

The following code is not compatible with Windows:

```
      - if: >
          !github.event.pull_request.draft && !(
            (github.actor == 'asyncapi-bot' && (
              startsWith(github.event.pull_request.title, 'ci: update of files from global .github repo') || 
              startsWith(github.event.pull_request.title, 'chore(release):')
            )) ||
            (github.actor == 'asyncapi-bot-eve' && (
              startsWith(github.event.pull_request.title, 'ci: update of files from global .github repo') || 
              startsWith(github.event.pull_request.title, 'chore(release):')
            )) ||
            (github.actor == 'allcontributors[bot]' && 
              startsWith(github.event.pull_request.title, 'docs: add')
            )
          )
        id: should_run
        name: Should Run
        run: echo "shouldrun=true" >> $GITHUB_OUTPUT
```

More specifically, I think the problem lies with `run: echo "shouldrun=true" >> $GITHUB_OUTPUT`.

We have two options as I see it.
1. We change the `run` command to be compatible with any system
2. We remove the check entirely.

Not sure which is the best approach.

derberg:
thanks for reporting

we cannot remove the check as then we are back in problem that PR tests run on every single PR, even these from bots where PR test is not needed - so yeah - our CI will cry

we need to improve check and basically have different approach if os is windows

looks like in general a known issue -> https://github.com/actions/runner/issues/2224

most voted solution is https://github.com/actions/runner/issues/2224 but to be honest adding `shell: bash` sounds best for me, as I understand it means that `bash` is then enforced to use also in windows.

@jonaslagoni the best approach would be if you could apply the fix first in your repo, and if it works there, then we can push fix through here? thoughts?

jonaslagoni:
@derberg seems like it works: https://github.com/asyncapi/modelina/actions/runs/8450691764/job/23147412309?pr=1912

derberg:
oh, yeah, thanks

I also noticed we already use this approach in another step https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L48

PR #280, opened 2024-04-02
ci: fix windows issues in few workflows

derberg:
**Description**

- `shell: bash`  resolves https://github.com/asyncapi/.github/issues/278
- I also removed the `npm install` with verbose as it was slowing down CI and with no benefits really. Now we do proper `npm ci`
- one more fix for windows cache (check inline comment

**Related issue(s)**

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -85,6 +85,7 @@ jobs:
       - name: Check if Node.js project and has package.json
         id: packagejson
         run: test -e ./package.json && echo "exists=true" >> $GITHUB_OUTPUT || echo "exists=false" >> $GITHUB_OUTPUT
+        shell: bash
       - if: steps.packagejson.outputs.exists == 'true'
         name: Check package-lock version
         uses: asyncapi/.github/.github/actions/get-node-version-from-package-lock@master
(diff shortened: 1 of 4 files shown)


.github/workflows/if-nodejs-release.yml
```diff
@@ -98,7 +99,14 @@ jobs:
...
         name: Install dependencies
-        run: npm install
+        id: first-installation
+        run: npm ci
+        continue-on-error: true
+      - if: steps.first-installation.outputs.status == 'failure' && steps.packagejson.outputs.exists == 'true'
+        name: Clear NPM cache and install deps again
+        run: | 
+          npm cache clean --force
+          npm ci
```
derberg, line 109: this is fixing https://github.com/asyncapi/cli/actions/runs/8520415954/job/23336484803

these issues keep repeating, we had this fix for PR testing workflow already

but issue back again https://github.com/npm/cli/issues/7308 and now blocking release

peter-rr: Any chance to reproduce the error locally? Or just possible on Github?

derberg: that happens only in windows, on local you could probably try to reproduce by doing the same as described in the error, manually damage local cache

KhudaDad414 (approved):
LGTM

.github/workflows/if-nodejs-release.yml
KhudaDad414: Hey @derberg, we've hit a classic chicken-and-egg problem with the CLI. Can't merge the PR due to a broken cache, but we can't fix the cache without releasing a new version... which we can't do without merging the PR. 

Should we wait for a bot to bump a package and fix things (since bot PRs are not tested), or do we need to patch the testing workflow too?

PR #281, opened 2024-04-03
ci: fix reinstallation of failure in release and pr testing

derberg:
After the fix from yesterday I noticed that the reinstallation that we had in PR testing actually never worked as we used wrong keywords

we should use `outcome` as describe in https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context


peter-rr (approved):
LGTM 👍

KhudaDad414 (approved):

PR #287, opened 2024-04-09
ci: improve node testing and release

derberg:
- changes to PR testing were battle-tested in https://github.com/asyncapi/cli/pull/1340 first
  - we had to apply workaround for npm version on windows
  - now we have a bit faster workflow as linting and assets gen is done on ubuntu only 
- changes to release workflow were battle-tested after merging https://github.com/asyncapi/cli/pull/1338 and seeing release passed without issues https://github.com/asyncapi/cli/actions/runs/8614265053 and we have new release https://github.com/asyncapi/cli/releases/tag/v1.8.0


KhudaDad414 (approved):
LGTM

PR #294, opened 2024-04-26
ci: add support for `next` release branch

jonaslagoni:
**Description**
In Modelina we have been using `next` since the beginning, however, we decided to remove it from the official workflow 🤨 We did a patch at some point in Modelina, and then I forgot about it.

This PR adds it officially.

### DIFF
--- a/.github/workflows/.releaserc
+++ b/.github/workflows/.releaserc
@@ -13,6 +13,8 @@ branches:
   prerelease: true
 - name: alpha
   prerelease: true
+- name: next
+  prerelease: true
 plugins:
 - - "@semantic-release/commit-analyzer"
   - preset: conventionalcommits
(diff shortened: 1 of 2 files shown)


derberg:
don't you use also a global release config for sematic-release? 
we need change in https://github.com/asyncapi/.github/blob/master/.github/workflows/.releaserc as well

jonaslagoni:
We dont no.

jonaslagoni:
In the next branch we use a much more customized version: https://github.com/asyncapi/modelina/blob/next/.releaserc

jonaslagoni:
I wonder if we should just remove Modelina from getting global CI change pushes...

derberg:
@jonaslagoni as long as you do not provide modifications in release workflow, I do not recommend it.

release workflows are separated from release config with topics `get-global-node-release-workflows` and `get-global-releaserc` so that is fine and you have a flexibility you need at the moment. Just please update in this PR the release config, so semantic-release for others knows what to do with merges to `next`

jonaslagoni:
@derberg done :v:

derberg (approved):

PR #314, opened 2024-09-11
ci: add command for transferring issue

### LINKED ISSUE #311: [FEATURE] Would be nice to have bot be able to move issues between repositories
### Why do we need this improvement?

To move a issue from one repo to another, you need to be codeowner to both the repositories that you are trying to move. Which is however not possible, so a bot command that does that would be awesome. 

Issues like https://github.com/asyncapi/cli/issues/1488 cannot be transfered because there is no common codeowner between CLI and modelina. 

### How will this change help?

Repositories that implement other AsyncAPI tools like CLI. When encounterd bug most of the time it has to do with something with the underlying library. So to be able to move the issue to the right repo, helps in solving the problem faster. 

### Screenshots

_No response_

### How could it be implemented/designed?

A simple bot command like `/rtm` that accepts the repo name to move as a input. Something like `/mv modelina`. Would be nice if only code owner of the repo could run this command. 

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

Souvikns:
**Description**
In this PR I am adding a new workflow file to trasfer issue from one respository to another in AsyncAPI organisation. The command is `/trasfer-issue {repo-name}` or `/ti {repo-name}` for example

```
/trasfer-issue cli
/ti bundler
```

**Related issue(s)**

Resolves #311


.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,35 @@
...
+      - created
+
+
+jobs:
+  transfer:
+    if: ${{(!github.event.issue.pull_request && github.event.issue.state != 'closed' && github.actor != 'asyncapi-bot') && (contains(github.event.comment.body, '/trasfer-issue') || contains(github.event.comment.body, '/ti'))}}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
derberg, line 16: use latest, which is I think v4, you need to check

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  issue_comment:
+    types:
+      - created
+
+
+jobs:
+  transfer:
+    if: ${{(!github.event.issue.pull_request && github.event.issue.state != 'closed' && github.actor != 'asyncapi-bot') && (contains(github.event.comment.body, '/trasfer-issue') || contains(github.event.comment.body, '/ti'))}}
```
derberg, line 12: we should probably start using starts with instead of contains

@KhudaDad414 thoughts?

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install GitHub CLI
+        run: |
+          sudo apt-get update 
+          sudo apt-get install gh
```
derberg, line 20: are you sure it must be installed, afaik runners have it by default. Check other workflows we have, they use `gh` without installation

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Authenticate GitHub CLI
+        run: |
+          gh action login --with-token <<< "${{ secrets.GITHUB_TOKEN }}"
```
derberg, line 23: are you sure it is needed? isn't gh preauthenticated already? also we should use GH_TOKEN secret if this is needed explicitly

derberg (commented):
left some comments, 

you also need to add this new workflow to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L141 to have it replicated to all other repos

.github/workflows/transfer-issue.yml
Souvikns: I found this https://github.com/marketplace/actions/setup-the-github-cli

Souvikns: I don't think we have any workflow that uses gh cli

.github/workflows/transfer-issue.yml
derberg: @Souvikns 

example of gh cli usage: https://github.com/asyncapi/spec/blob/master/.github/workflows/update-spec.yaml#L61

some more docs: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows

basically no login and installation is necessary

Souvikns: Alright, I removed the installation and login, so now I only check parse the comment body and then and use it to trasfer issues.

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,25 @@
...
+name: Transfer Issues between repositories
+
+on:
+  issue_comment:
+    types:
+      - created
+
+jobs:
+  transfer:
+    if: ${{(!github.event.issue.pull_request && github.event.issue.state != 'closed' && github.actor != 'asyncapi-bot') && (startsWith(github.event.comment.body, '/trasfer-issue') || startsWith(github.event.comment.body, '/ti'))}}
```
derberg, line 11:
```suggestion
    if: ${{(!github.event.issue.pull_request && github.event.issue.state != 'closed' && github.actor != 'asyncapi-bot') && (startsWith(github.event.comment.body, '/transfer-issue') || startsWith(github.event.comment.body, '/ti'))}}
```

👁️

.github/workflows/help-command.yml
```diff
@@ -58,5 +58,6 @@ jobs:
               At the moment the following comments are supported in issues:
   
               - \`/good-first-issue {js | ts | java | go | docs | design | ci-cd}\` or \`/gfi {js | ts | java | go | docs | design | ci-cd}\` - label an issue as a \`good first issue\`.
-              example: \`/gfi js\` or \`/good-first-issue ci-cd\``
+              example: \`/gfi js\` or \`/good-first-issue ci-cd\`
+              - \`/trasfer-issue {repo-name}\` or \`/ti {repo-name}\` - trasfers issue from the source repository to the other repository passed by the user. example: \`/ti cli\` or \`/trasfer-issue cli\`
```
derberg, line 62: `trasfer` all over the place

derberg: also aren't we missing a closing ` at the end of the line here?

would be good if you test new workflow and this help workflow in private repo to be 100% sure it all works - before we push it to all the other repos

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Extract Input
+        id: extract_step
+        run: |
+          COMMENT="${{github.event.comment.body}}"
+          REPO=$(echo $COMMENT | awk '{print $2}')
+          echo repo=$REPO >> $GITHUB_OUTPUT
+      - name: Trasfer Issue
```
derberg, line 22: would be very useful to have some interaction with user of the command here, 

if transfer fails (most probably some typo of repo name of misuse of command) we should drop a proper comment in issue that command failed, and they have to make it better - I'm pretty sure we do it for good first issue command

Souvikns: I have added a step to check if the repo is under the same org and write a comment explaining this to the user.

.github/workflows/help-command.yml
Souvikns: Don't think any ` is missing, it is rendering this 

- `/transfer-issue {repo-name}` or `/ti {repo-name}` - transfer issue from the source repository to the other repository passed by the user. example: `/ti cli` or `/transfer-issue cli`

.github/workflows/help-command.yml
```diff
@@ -58,5 +58,6 @@ jobs:
               At the moment the following comments are supported in issues:
   
               - \`/good-first-issue {js | ts | java | go | docs | design | ci-cd}\` or \`/gfi {js | ts | java | go | docs | design | ci-cd}\` - label an issue as a \`good first issue\`.
-              example: \`/gfi js\` or \`/good-first-issue ci-cd\``
+              example: \`/gfi js\` or \`/good-first-issue ci-cd\`
+              - \`/transfer-issue {repo-name}\` or \`/ti {repo-name}\` - transfer issue from the source repository to the other repository passed by the user. example: \`/ti cli\` or \`/transfer-issue cli\`.
```
derberg, line 62: look at below what is missing in my opinion

Screenshot 2024-11-11 at 11 14 06

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,53 @@
...
+          echo repo=$REPO >> $GITHUB_OUTPUT
+      - name: Check Repo
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const r = process.env.GITHUB_REPOSITORY
+            const [owner, repo] = r.split('/')
+            const repoToMove = process.env.REPO_TO_MOVE
+            const issue_number = process.env.ISSUE_NUMBER
```
derberg, line 30: please always use `context` if code is executed in `github-script` action, so do not read for example issue number from env variables but like this: `context.issue.number`

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,53 @@
...
+                issue_number,
+                body
+              })
+                process.exit(1)
+            }
+      - name: Transfer Issue
+        id: transferIssue
+        working-directory: ./
+        run: |
+          gh issue transfer ${{github.event.issue.number}} asyncapi/${{steps.extract_step.outputs.repo}}
```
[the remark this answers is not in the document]
Souvikns: That is basically what I am doing, I am not starting the transfer process unless I can ensure that the repo is correct and under the AsyncAPI organization.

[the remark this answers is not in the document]
Souvikns: I think a better error message could be nice. Also, I don't see any way the transfer command failing if the check step is passed.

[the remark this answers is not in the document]
derberg: right, you verify if repo exists using api call, that should be enough

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,53 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
derberg, line 52:
```suggestion
          GH_TOKEN: ${{ secrets.GH_TOKEN }}
```

changing to out bot token

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,53 @@
...
+          REPO=$(echo $COMMENT | awk '{print $2}')
+          echo repo=$REPO >> $GITHUB_OUTPUT
+      - name: Check Repo
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const r = process.env.GITHUB_REPOSITORY
+            const [owner, repo] = r.split('/')
+            const repoToMove = process.env.REPO_TO_MOVE
```
derberg, line 29: technically where is `REPO_TO_MOVE` and `GITHUB_REPOSITORY` coming from?

derberg:
@Souvikns ping in case you forgot about this one

.github/workflows/transfer-issue.yml
Souvikns: `GITHUB_REPOSITORY` is a env variable that is already present in the running it returns you repository name like `asyncapi/cli` format. 

`REPO_TO_MOVE` is something I am extracting in the `Extract Input` step and storing it in the env, and while calling the step `Check Repo` I am passing `REPO_TO_MOVE` as a env on line 55.

Souvikns:
I have tested the action here: https://github.com/Klifu/API/actions/runs/12532481994. 
You can play around with the functionalities using this repo: https://github.com/Klifu/API, for now only `CLI` repo has issues so you can transfer issues to there.

.github/workflows/transfer-issue.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          COMMENT="${{github.event.comment.body}}"
+          REPO=$(echo $COMMENT | awk '{print $2}')
+          echo repo=$REPO >> $GITHUB_OUTPUT
+      - name: Check Repo
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const r = process.env.GITHUB_REPOSITORY
```
derberg, line 27: I suggest you switch to use standard way of doing things accessing context: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

`${{ github.repository }}`

.github/workflows/transfer-issue.yml
derberg: there is no line `55` and also in `Extract Input` there is no such thing as `REPO_TO_MOVE` but `repo`

and anyway you should access it like `steps.extract_step.outputs. repo`

derberg (changes requested):

Souvikns:
@derberg I made the changes, now fetching repo from github context and added missing env for `REPO_TO_MOVE`.

derberg:
code wise lgtm, did you test it in some repo? just to be 100% sure? as once I merge this one, it will be pushed to all repos.

maybe you can add it first to CLI manually, with the same file name (so it will be later overwritten by this PR))

derberg:
I'm only referring to `transfer-issue.yml` file

also please add the same header comment as other workflows here have

derberg:
@Souvikns you need help?

Souvikns:
> code wise lgtm, did you test it in some repo? just to be 100% sure? as once I merge this one, it will be pushed to all repos.
> 
> maybe you can add it first to CLI manually, with the same file name (so it will be later overwritten by this PR))

I tested in a org that I have, and it is working as expected

Souvikns:
https://github.com/Klifu/API/actions/runs/13002693693/job/36264188187 I tested in one my org and it is working as expected

derberg (approved):
lgtm

derberg:
I'll marge in few hours when most of us sleep as it will trigger CI in all repos, and yeah, we don't want to stress CI/CD setup now

PR #317, opened 2024-10-09
fix: add destination for .prettierignore replication

### LINKED ISSUE #316: add `destination` for prettierignore config replication
In https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L211 we need one more property called `destination` that is pointing to `./` as for now replicator moves prettier ignore not to the root but inside workflows dir. Example https://github.com/asyncapi/generator/pull/1289

SaxenaAnushka102:
**Description**

- Added the `destination` property to the `replicate_prettierignore_file` step in the `.github/workflows/global-replicator.yml` workflow.
- This ensures that the `.prettierignore` file is replicated correctly in the repository's root directory (`./`).

**Related issue(s)**

- Fixes #316


derberg (approved):
Thank you!

PR #327, opened 2025-02-10
ci: fix co-authors with pagination

### LINKED ISSUE #263: Fix wrong format for Co-authored automerged commits + pagination
#### Describe the bug

Commit Co-authoring allow us to merge PR's and tell Github who were the committers so it does not appear only the PR creator as the author of all the code on it. Unfortunately, it doesn't work at this moment due to a bug in our side. 
The way it works is described [here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).
We basically make it happen in [this action](https://github.com/asyncapi/.github/blob/master/.github/workflows/automerge-for-humans-merging.yml#L24).

However, the final commit message contains a wrong format that makes the co-authoring feature fail.

It seems like a format problem: no new line character is being printed (each co-author should appear in new line). Instead, the new line characters appears encoded (`%0A`) :disappointed:

#### How to Reproduce
See an example here: https://github.com/asyncapi/parser-js/commit/4799f2a1f5fd538e16f28a0490a06a6fe89caee1

#### Expected behavior
Each co-author line (`Co-authored-by: ...`) should appear in a new line as the documentation says. 
Additionally, we should find an alternative to curl + jq or maybe use some bash so we can implement pagination and by pass the limitation of 100 commit results. There are PR's that have more than 100 commits, specially in long-live branches.

Shurtu-gal:
**Description**

- Fixed coauthor list not appearing properly.
- Added proper pagination for larger PRs.

**Workflow example**
https://github.com/Shurtu-gal/action-test-bed/actions/runs/13243968167/job/36965817715

**Related issue(s)**
Fixes #263


derberg (approved):
great work!

Shurtu-gal:
Workflow run for future ref: https://github.com/Shurtu-gal/action-test-bed/actions/runs/13367967898/job/37329935224#step:4:1

PR #340, opened 2025-04-25
ci: update global files 

thulieblack:
add Holopin yml and update the global file

### DIFF
--- a/.github/workflows/global-replicator.yml
+++ b/.github/workflows/global-replicator.yml
@@ -237,3 +237,25 @@ jobs:
           committer_email: info@asyncapi.io
           commit_message: "ci: update update-maintainers-trigger.yaml workflow from global .github repo"
           bot_branch_name: bot/update-files-from-global-repo
+
+# The holopin.yml file is replicated from the global .github repository to ensure consistency across all repositories.
+# This file is used to manage Holopin badge configurations for contributors, maintainers, and other roles in the organization.
+  replicate_holopin_file:
+    if: startsWith(github.repository, 'asyncapi/')
+    name: Replicate holopin.yml file to repositories
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Replicating holopin.yml file
+        uses: derberg/manage-files-in-multiple-repositories@beecbe897cf5ed7f3de5a791a3f2d70102fe7c25
+        with:
+          github_token: ${{ secrets.GH_TOKEN }}
+          patterns_to_include: .github/holopin.yml
+          topics_to_include: get-global-holopin
+          committer_username: asyncapi-bot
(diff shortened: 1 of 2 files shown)


.github/workflows/global-replicator.yml
```diff
@@ -237,3 +237,25 @@ jobs:
...
+    name: Replicate holopin.yml file to repositories
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Replicating holopin.yml file
+        uses: derberg/manage-files-in-multiple-repositories@beecbe897cf5ed7f3de5a791a3f2d70102fe7c25
+        with:
+          github_token: ${{ secrets.GH_TOKEN }}
+          patterns_to_include: .github/holopin.yml
```
derberg, line 254: shouldn't this be located in the root of the repo? like it is for example with `CONTRIBUTING.md` ?

.github/workflows/global-replicator.yml
```diff
@@ -237,3 +237,25 @@ jobs:
...
+          topics_to_include: get-global-holopin
```
derberg, line 255: so not every repo gets it by default?

thulieblack:
okay update @derberg

holopin.yml
```diff
@@ -0,0 +1,34 @@
...
+    alias: silver-badge
+  - id: cmae2583o30150dl56bl3ms9z
+    alias: gold-badge
+  - id: cmae26orl110420dkypr5dy0yn
+    alias: platinum-badge
+  - id: cmae288m9116470dky1ku70j1u
+    alias: diamond-badge
+ # more stickers
+
+# To issue the badges tag @holopin-bot @username sticker-alias
```
derberg, line 34: ia there already some document that we can also reference here?

derberg: `To issue the badges tag @holopin-bot @username sticker-alias` bothers me - basically how I can assign sponsor badges to sponsors

thulieblack: It will be like @holopin-bot @derberg bronze badge 

The docs we have is this https://docs.holopin.io/integrations/github

thulieblack: @derberg

thulieblack: Oooh there are many ways to issue the badges apart from the above https://docs.holopin.io/issuing-rewards/regular-badges

derberg: cool!

so the only thing left where we can reference people form inside `holopin.yml`. I imagine many maintainers will not know what it is as not all follow community news regularly. So they need to open the file and see quick reference why it landed in their repo and what they can do with it

derberg: please do

thulieblack:
your review @derberg

derberg (approved):

PR #343, opened 2025-05-28
docs: add comments to all jobs in `global-replicator`

Aahil13:


derberg (commented):
looking great, just update with latest master, there is new job added by @thulieblack that requires you to add one more comment in yaml

Issue #344, opened 2025-06-03
[BUG] Transfer Issue workflow not working.


Souvikns:
### Describe the bug.

I was trying to transfer a issue from CLI, to modelina and used the command `/ti modelina` it did not work and it looks like the github token does not have the required permission 

https://github.com/asyncapi/cli/actions/runs/15362319072/job/43230998207 Adding the link to the workflow where the command failed. 

### Expected behavior

Should be able to transfer issues from one repo to another. 

### Screenshots

https://github.com/asyncapi/cli/actions/runs/15362319072/job/43230998207

### How to Reproduce

1. I first did this
2. I then did this
3. And so on . . . 

https://github.com/asyncapi/cli/actions/runs/15362319072/job/43230998207 here is the link check out a failed workflow, but to reproduce you can comment on any issue like `/ti <repo-name-to-trasnfer>`. 

### 🖥️ Device Information [optional]

- Operating System (OS): Linux
- Browser: Firefox
- Browser Version: 139.0.1

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

Souvikns:
@derberg I think we just need to update token with access to this particular thing.

PR #376, opened 2025-12-17
fix: update transfer-issue.yml

spc-28:
**Description**

Addresses `/ti` command permission errors by using `secrets.GH_TOKEN` instead of `secrets.GITHUB_TOKEN` (limited to current repo only).

Issue: #344


Shurtu-gal:
@spc-28 this might work but it was done like this in https://github.com/asyncapi/.github/pull/342 probably due to security reason or due to it not working with GH_TOKEN as well.

Shurtu-gal:
I don't understand how does that manage to make this more secure?

derberg:
Closing for now

PR #410, opened 2026-07-30
ci: bump all actions to latest, pin to commit SHAs with version comments

derberg:
All third-party actions were already SHA-pinned, but most pins were several majors behind and the version comments were stale or missing. This bumps every action to its latest release, keeps the full commit SHA, and normalizes the trailing comment to `# vX.Y.Z`.

For every major bump I checked the inputs we pass against the new version's `action.yml` — all still exist. Extra checks on the risky ones:

- `actions/checkout` v4 → v7.0.1: `persist-credentials` still defaults to `true`. The new `allow-unsafe-pr-checkout` gate only blocks fork PR-head checkouts in `pull_request_target` — none of our workflows do that (we only check out the base repo).
- `actions/github-script` v7 → v9: all inline scripts already use `github.rest.*`, so no API changes needed.
- `golangci/golangci-lint-action` v6.5.1 → v9.3.0: we pass no inputs, so the action now installs latest golangci-lint **v2.x**. Go repos still on v1 config format will need to migrate their `.golangci.yml` (https://golangci-lint.run/product/migration-guide/). Updated the stale inline comments accordingly.
- `pascalgn/automerge-action` v0.15.6 → v0.16.4: all `MERGE_*` env vars still supported. `GITHUB_LOGIN` in automerge.yml turns out to be inert in both old and new versions (never read by the action); left as is.
- `peter-evans/create-pull-request` v4.2.4/v7.0.8 → v8.1.1: unified both pins, all inputs (`token`, `commit-message`, `committer`, `author`, `title`, `body`, `branch`) unchanged.
- `actions/stale` v9.1.0 → v11.0.0: all 10 inputs we use still exist.

Already at latest, only comments added/fixed: `chinthakagodawita/autoupdate` (v1.7.0), `derberg/manage-files-in-multiple-repositories` (v2), `derberg/npm-dependency-manager-for-your-github-org` (v6.2.0), `m1ner79/Github-Twittction` (v2.0.0).

Left alone: internal `asyncapi/.github/.github/actions/*@master` refs (intentional, marked NOSONAR), the pinned zizmor CLI version, and the pinned `npm@8.19.4` installs.

Validation: `actionlint` clean, `zizmor` (auditor persona, repo config) reports the same single pre-existing low-confidence finding as master — nothing new introduced.

### DIFF
--- a/.github/workflows/add-good-first-issue-labels.yml
+++ b/.github/workflows/add-good-first-issue-labels.yml
@@ -20,7 +20,7 @@ jobs:
       issues: write    # This is needed to add labels to issues.
     steps:
       - name: Add label
-        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           github-token: ${{ github.token }}
           script: |
--- a/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
+++ b/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Add ready-to-merge label
-        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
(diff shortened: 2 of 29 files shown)


derberg:
`gh pr create` is not the only thing that we would have to do as peter also handles staging, has build in functionality to not push and create if no changes - zizmore is wrong

Shurtu-gal (approved):

